PR #151, opened 2023-02-13
Some very minor makefile corrections

iahunq:
I discovered them when trying to have Objeck build on MINGW64 (MSVCRT, not UCRT64). Finally I was able to build it and it worked well. The only thing depends on UCRT64 on Objeck code is `_dupenv_s` that doesn't available on MSVCRT, replace it with the plain old `getenv` and it compiled and worked (Unicode strings printed on the screen just fine). I decided to not upstream it because I have trouble with the macros logic. It's impossible for me to use macros to have it build on both MSVCRT and UCRT64. It will only build on MSVCRT if I replaced `_dupenv_s` by `getenv` completely. After all it's a just an adventure of me and it's not worth to do so as `getenv` is deprecated as it's neither secure nor thread-safe and UCRT64 is superior to MSVCRT. But the only thing I know now is I could hack Objeck to build on MINGW64 (MSVCRT). I will not go any further. It's enough work for today ^^


iahunq:
Please merge it @objeck. It's a very trivial PR so nothing could go wrong. I will never do such a thing like adding Cygwin again :(

objeck:
Merged. Over the years, I have reverted many commits; no worries.

PR #152, opened 2023-02-14
Remove leftover Cygwin makefile

iahunq:


objeck:
Thanks, I will remove ``Makefile.cygwin.amd64`` file. The ``I386_PGBYTES``definition should be removed, it is no longer needed.

PR #209, opened 2023-04-11
copy dlls

iahung:
truly standalone, no longer need to put ucrt64/bin or clang64/bin to PATH


iahung:
> Nice, thank you! What happens if the version numbers change?

Version numbers of what? The dlls? It copies the dlls from the MSYS2 installation whenever you run the deploy script so the dlls will always be the latest version if your MSYS2 installation is up to date.

objeck:
Yes, the version numbers of MSYS2 DLLs. If one of the libraries is upgraded is it still copied.

iahung:
> If one of the libraries is upgraded is it still copied.

What do you mean? It will copy the dlls at the time the deploy script is run. After that, if the libraries are updated and thus the dlls change, you will need to run the deploy script again to copy the updated dlls. This also means a rebuild of Objeck even if the source code has not changed.

PR #416, opened 2023-12-10
Objeck syntax coloring for Geany

tqo50396:
Note: Geany doesn't support highlighting of Objeck style comments. Use C style comments instead.


objeck:
@thundervox Thank you!

objeck:
Oops, I meant @tqo50396

PR #517, opened 2025-07-29
Update lame.cpp

Curculigo:
Fix compilation error on Debian 12.


Curculigo:
The change is too trivial so I think the developer can commit it themselves.

PR #614, opened 2026-08-09
HTTP/3: enable on Windows (WinHTTP/MsQuic), check CSPRNG results, enable linux-arm64

objeck:
Three small, independent HTTP/3 fixes surfaced while planning the Windows port (#613).

## Security — QUIC connection IDs could be uninitialised stack memory

All five `gnutls_rnd` calls ignored their return value, and the initial connection IDs were declared **uninitialised**. `gnutls_rnd` leaves the buffer untouched on failure, so a CSPRNG failure transmitted whatever was on the stack in the **cleartext QUIC Initial header** — a memory disclosure plus a trivially predictable connection ID, and at the reset-token site a **forgeable stateless reset** letting an off-path attacker tear down connections.

Every site now goes through a checked `h3_random_bytes`; IDs are zero-initialised; connection setup fails closed; and `h3_get_new_connection_id_cb` returns `NGTCP2_ERR_CALLBACK_FAILURE` instead of proceeding. Secrets draw from `GNUTLS_RND_KEY`. **This affects Linux and macOS today** — unrelated to Windows.

## Platform — HTTP/3 was silently unavailable on Linux ARM64

`Makefile.arm64` never defined `OBJECK_HAS_NGTCP2` and linked no ngtcp2/nghttp3/gnutls, so `Http3Client` had no engine there — while CI already installs those deps on that leg (`ci-build.yml:868`) and throws them away. One-line enable; the dependencies were already paid for.

## Docs — stop the false promise now

`net_quic.obs` and the CHANGELOG now state the real matrix: **Linux x64/ARM64 and macOS; not Windows** (use `HttpClient`/HTTP2 there). `Http3Client` being documented without qualification while doing nothing on Windows was the original complaint — this corrects it immediately rather than waiting for the port.

## Verification

`net_quic.obs` recompiles clean. The ngtcp2 code cannot build on Windows, so **CI linux-x64 — and now linux-arm64 — is the compiler check**; linux-arm64 compiling this code for the first time is itself meaningful signal.


objeck:
Added a third commit (`ab40014bb`) — same theme, so it belongs here rather than in a stacked PR.

The docs commit states the platform matrix **in prose**. This makes the same truth **machine-readable**: `runtime.feature.http3` / `runtime.feature.http2`.

The reason this is needed at all: the HTTP/3 traps are compiled unconditionally — an unsupported build still *has* an `Http3Connect`, its `#else` branch just fails. So "the API exists" proves nothing, and a caller could not distinguish **not compiled in** from **the network failed**. That ambiguity is exactly how `Http3Client` shipped documented-and-dead on Windows without anyone noticing.

Verified on Windows: `http2=1 http3=0`, matching the documented matrix exactly.

The new regression test asserts http2 is **never** 0 — every VM build config defines `OBJECK_HAS_NGHTTP2`, so a 0 means a build silently lost nghttp2, the same class of regression that hid the HTTP/3 gap. http3 is platform-dependent, so it is reported but deliberately not asserted. The test is negative-controlled: pointing the same assertions at `runtime.memory.used` makes them fail, so they aren't vacuous.

Note `TESTS.md` is large in the diff — regenerating it caught up **13 already-committed tests** that were never added to the manifest (172 → 186), most of them mine from recent work. Only one of those is new here.

objeck:
## HTTP/3 now works on Windows (`d6bdd03e8`)

This PR started as "document that HTTP/3 is POSIX-only." It now **fixes** it instead — the false promise is gone rather than merely labelled.

### Why not the obvious fix
The POSIX h3 TLS layer is hard-bound to GnuTLS (16 call sites + `ngtcp2_crypto_gnutls`) and Windows uses mbedTLS, so enabling the existing path meant GnuTLS-on-MSVC via vcpkg or a second TLS backend over ~700 lines — both large, both risking the working POSIX path.

Windows already ships an HTTP/3 stack. `core/vm/win_http3.h` sits behind `OBJECK_HAS_WINHTTP_H3`; the three traps gained `#elif` branches. **The POSIX path is untouched.** No vendored QUIC library, no second TLS backend, no new DLL — `winhttp.lib` ships with Windows. Needs Win11 / Server 2022+.

### The part that actually matters
WinHTTP treats HTTP/3 as a **preference** and falls back transparently, so a 200 proves nothing. Every request asserts `WINHTTP_OPTION_HTTP_PROTOCOL_USED` and **fails unless it really was HTTP/3** — a client that quietly spoke HTTP/2 would be the same false promise this fixes.

(Also non-obvious, cost a probe to find: `WINHTTP_OPTION_ENABLE_HTTP_PROTOCOL` must be set on the **session** handle *before* the request is created. Setting it later is accepted and silently ignored.)

### Verified live on Win11
| Host | Result |
|---|---|
| cloudflare-quic.com | 200, 125959 b |
| www.google.com | 200, 82391 b |
| quic.nginx.org | 200, 2523 b |
| blog.cloudflare.com | 200, 186913 b |
| example.com | **refused** (serves HTTP/1.1) |
| neverssl.com | **refused** |

Plus POST with body + content-type, custom headers, session reuse, and Close. The refusals are confirmed *guard working* rather than *failed to connect* — a WinHTTP probe shows those hosts really do downgrade.

**Full Windows regression suite: 187 passed, 0 failed.**

### Two things reviewers should weigh
1. **No CI coverage for this path.** The network step is `if: matrix.platform == 'linux-x64'` and non-gating, so nothing here runs the Windows backend in CI. All verification above is manual, on this box. Extending that step to windows-x64 is a real option — I did not do it, since a half-wired non-gating step is worse than a documented gap.
2. **msys2 Windows builds are deliberately excluded.** They have no engine and correctly report `runtime.feature.http3 == 0`. Enabling them is one define plus `-lwinhttp`, but mingw's `winhttp.h` is untested here and I would rather not break a build I cannot run.

The new regression test runs in the **gating** suite (`run_regression.sh` runs every `.obs` there), so it treats anything that could be the network as SKIP and reserves FAIL for the one thing network conditions cannot cause: a non-HTTP/3 host reporting success.

PR #628, opened 2026-08-10
fix(onnx): ONNX session creation was broken on Linux and MSYS2

objeck:
Verified in WSL2 against the vendored runtime — not inferred. **Three defects compounding into one silent failure.**

### 1. `AppendExecutionProvider("CUDA", …)` always threw

ORT's generic string form doesn't accept CUDA — it recognises only OPENVINO, SNPE, XNNPACK, QNN, WEBNN and AZURE. CUDA needs the typed entry point. **The thrown message doesn't even mention CUDA**, which is why this was never diagnosed:

```
Unknown provider name. Currently supported values are 'OPENVINO', 'SNPE',
'XNNPACK', 'QNN', 'WEBNN' and 'AZURE'
```

### 2. The vendored runtime is CPU-only

Despite living under `eq/cuda/lib`, `libonnxruntime.so.1.19.0` returns exactly `CPUExecutionProvider` from `GetAvailableProviders()`. So even the typed API couldn't have worked. Meanwhile `deploy_posix.sh` and both `deploy_msys2-*.sh` built with `./build.sh cuda` → `-DONNX_EP_CUDA`, guaranteeing the throw on **every** session creation.

### 3. It failed silently

The `catch` returns without `APITools_SetIntValue`, so callers got a **null session rather than an error**. That's why a total failure of ONNX on Linux went unnoticed.

## Fixes

- CUDA branch checks `GetAvailableProviders()` **before** appending, refuses with a message naming what's actually available, and uses the **typed** `AppendExecutionProvider_CUDA`. It refuses rather than falling back — a session that quietly ran on CPU when CUDA was asked for is indistinguishable from success.
- POSIX/MSYS2 deploys build `./build.sh cpu`, matching the CPU-only runtime they ship.
- `build.sh`'s **cpu mode linked a bare `-lonnxruntime` with no `-L`**, requiring a system-installed runtime and failing to link without one — even though a usable runtime sits in `cuda/lib/x64/lib`. Now uses the vendored one.

## Verified in WSL2 (Ubuntu 24.04)

| | result |
| --- | --- |
| **old** | THREW: *Unknown provider name. Currently supported values are 'OPENVINO', 'SNPE', 'XNNPACK', 'QNN', 'WEBNN' and 'AZURE'* |
| **new** | *CUDA requested but linked onnxruntime does not provide it. Available: CPUExecutionProvider* |

**Not verified:** a full native Linux build — `onnx.cpp` needs OpenCV headers, absent from this WSL2. The provider logic was exercised standalone against the real runtime; the Windows DML build compiles clean.

Note `libonnxruntime.so` / `.so.1` are checked out as 19- and 24-byte text files rather than symlinks (NTFS artifact) — harmless on a native Linux clone, but it breaks a WSL2 build over `/mnt/c`.

### DIFF
--- a/core/lib/onnx/eq/build.sh
+++ b/core/lib/onnx/eq/build.sh
@@ -33,7 +33,12 @@ case "$PROVIDER" in
 	cpu)
 		EP_DEFINE=""
 		ORT_INCLUDE="-I./cuda/lib/include"
-		ORT_LIB="-lonnxruntime"
+		# Use the VENDORED runtime, same as the cuda case. This previously said
+		# just -lonnxruntime, which requires a system-installed onnxruntime and
+		# fails to link on a machine that has none -- even though a usable
+		# runtime is sitting in cuda/lib/x64/lib. That runtime is CPU-only, so
+		# it is exactly the right one for this mode.
+		ORT_LIB="-L./cuda/lib/x64/lib -lonnxruntime"
 		;;
 	*)
 		echo "Unknown provider: $PROVIDER"
--- a/core/lib/onnx/eq/onnx.cpp
+++ b/core/lib/onnx/eq/onnx.cpp
@@ -82,6 +82,39 @@ extern "C" {
 #if defined(ONNX_EP_DML)
          const std::string compiled_ep = "dml";
 #elif defined(ONNX_EP_CUDA)
+         const std::string compiled_ep = "cuda";
+#elif defined(ONNX_EP_QNN)
+         const std::string compiled_ep = "qnn";
+#elif defined(ONNX_EP_VITIS)
+         const std::string compiled_ep = "vitisai";
+#elif defined(ONNX_EP_COREML)
+         const std::string compiled_ep = "coreml";
+#else
+         const std::string compiled_ep = "cpu";
+#endif
+
+         const bool use_cpu = (requested_ep == "cpu");
+         if(!requested_ep.empty() && !use_cpu && requested_ep != compiled_ep) {
+            std::wcerr << L">>> ONNX: execution provider '" << BytesToUnicode(requested_ep)
+                       << L"' was requested, but this build provides '"
+                       << BytesToUnicode(compiled_ep)
+                       << L"'. Use ep=" << BytesToUnicode(compiled_ep)
+                       << L" or ep=cpu. <<<" << std::endl;
+            return;
+         }
+
+         Ort::SessionOptions session_options;
+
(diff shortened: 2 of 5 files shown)


objeck:
**Update — the native Linux build is now verified end to end.**

The PR said a full native build was unverified because OpenCV headers appeared to be missing. That was my error: OpenCV 4.6.0 **was** present, and `build.sh` locates it correctly through `pkg-config`. My manual compile had simply omitted the pkg-config flags.

With that corrected, in WSL2 (Ubuntu 24.04):

- `onnx.cpp` **compiles clean** with these changes
- it **links** into a 249 KB `libobjk_onnx.so`

So the fix is verified on the platform it repairs, not just on Windows.

One genuine obstacle remains, and it is environmental rather than a code defect: `eq/cuda/lib/x64/lib/libonnxruntime.so` and `.so.1` are checked out as **19- and 24-byte text files instead of symlinks** (NTFS doesn't preserve them). `ld` then reports *"file format not recognized; treating as linker script"*. This is harmless on a native Linux clone, but it will stop any WSL2 build over `/mnt/c`. I linked against a directory with the symlinks restored to prove the link succeeds.

Worth considering separately: `build.sh` could detect that case and emit a clear message, since the raw `ld` error points nowhere useful.

objeck:
**Fully verified on native Linux — and doing so caught a real bug in this PR.**

Building on a native clone (intact symlinks, unlike a WSL2 build over `/mnt/c`):

| mode | result |
| --- | --- |
| `./build.sh cpu` | **OK** — 249,440 byte `libobjk_onnx.so` |
| `./build.sh cuda` | **OK** — 254,440 byte `libobjk_onnx.so` |

### The bug this caught

The first version of the CUDA guard **did not compile**:

```
onnx.cpp:126: 'session_options' was not declared in this scope
onnx.cpp:143: 'compiled_ep' was not declared in this scope
```

My splice had targeted the *first* `#elif defined(ONNX_EP_CUDA)` in the file, which — after the `ep` change added its own provider chain — was the `compiled_ep` **declaration** chain rather than the provider-**append** chain. The session code landed before `session_options` existed.

**The Windows build could not have caught this.** It compiles only the DML arm, so the CUDA branch is never seen. That's the same structural blind spot as the ARM64 JIT fix earlier today, where only the ARM64 legs exercised the changed code.

Fixed in `01b30f2f8e`'s follow-up: declaration arm restored, guard moved to the append chain, both modes now build.

PR #632, opened 2026-08-10
feat(repl): run F5 as a subprocess, and round out the editor

objeck:
Implements the approved plan (`.claude/plans/eventual-floating-crane.md`). Branched off the #631 revert, so it does **not** inherit the worker thread.

## Part A — F5 as a subprocess

F5 ran in-process on the UI thread, so any loop froze obi. A worker thread was *worse* — it hung even `7->PrintLine()`, because the in-process VM has **main-thread affinity** (process-global `Loader`/`MemoryManager`/`StackInterpreter` plus its own GC threads). It can't leave the UI thread, so the run had to leave the **process**.

- **`child_run.h`** (new) — `Tui::Child`. Windows: `CreateProcessW` + `CreatePipe` + `PeekNamedPipe` + `TerminateProcess`. POSIX: `fork`/`execvp`/`pipe`, `O_NONBLOCK`, `SIGTERM`→`SIGKILL`. Argv vectors, **no shell** — deliberate, since the path originates in a user buffer.
- **`editor.cpp`** — `ExecutableDir()` resolves `obc`/`obr` as siblings of obi. F5 now returns a `RunPlan` (argv + `OBJECK_LIB_PATH`): buffer → temp `.obs`, then `obc`, then `obr`. `-lib` gets `compiler_libs` **bare** — prepending `lang.obl,` is right for `ObjeckLang` and wrong for `obc`, which filters `lang`/`collect` itself.
- **`DoRun`** — two-phase pump: drain child → pane, `Draw()`, `ReadKey()` each ~100 ms tick. **Esc kills the child.** No terminal change needed; `ReadKey` already timed out on both platforms.
- **`IoCapture` dropped from F5.** The child owns its pipes, and a pipe fails `GetConsoleMode`, so the VM falls back to `wcout` instead of `WriteConsoleW` — the Windows console bypass that swallowed output is fixed **by construction**.

## Part B — editor quality

| item | before → after |
|---|---|
| Smart return | no auto-indent at all → inherits indent, +1 level after `{` |
| Highlighting | none → keyword/number/string via the **already-linked** compiler `Scanner`, cached, composed *under* selection |
| vi mode | confirmed present (F2) → adds `w` `b` `A` `I` `O` `D` |
| Linewise paste | **bypassed the read-only guard** — could inject lines into the class/function frame → guarded |
| `want_col` | character index → display column (cursor no longer drifts over tabs/wide chars) |
| `Prompt` resize, `DoSave` `.obs` | ignored / unenforced → handled |

## Verified

Builds clean via `core/repl/repl.sln` (the single vcxproj fails on `objeck.lib`). I independently rebuilt and hand-reviewed the spawn paths, argv construction, and selection/highlight compositing.

## ⚠️ NOT verified — please read before merging

- **No interactive testing.** `/e` needs a real TTY; none of this has been exercised by a human yet.
- **The POSIX half of `child_run.h` has never been compiled** — this is a Windows box. It mirrors `obu.cpp`'s proven pattern, but needs a Linux/macOS build.
- **Do not auto-merge on CI green.** CI cannot exercise any of this. That is precisely how the worker-thread regression reached master.

**Manual test matrix:** `7->PrintLine()` → output in pane · compile error → F8 jumps · `while(true){};` + **Esc** → cancels, obi alive · Enter after `{` → indents · keywords/strings colored · F2 vi → `w`/`A`/`D` · linewise paste onto a frame line → refused.

## Deliberately skipped (documented in `EDITOR_DESIGN.md`)

Multi-line-string and comment coloring — the scanner reports those at the closing quote, so they're left **uncolored rather than mis-colored**. Also char literals, incremental highlight caching, full vi, OS clipboard, find/goto-line, mouse.

### DIFF
--- a/core/repl/EDITOR_DESIGN.md
+++ b/core/repl/EDITOR_DESIGN.md
@@ -1,5 +1,34 @@
 # Full-screen editor for `obi` — design
 
+> **STATUS (branch `feat/obi-editor-subprocess-and-quality`).** Phases 1–4 are
+> implemented. Two design points below are now superseded:
+>
+> - **Run is a subprocess, not in-process.** F5 saves the buffer to a temp
+>   `.obs`, compiles it with `obc`, then runs the `.obe` with `obr` — each a
+>   child process whose merged stdout/stderr the pane drains without blocking
+>   (`child_run.h`, `tui_editor.h::DoRun`). The in-process VM has main-thread
+>   affinity and would deadlock if driven off the UI thread, so "a run pane does
+>   not need a subprocess" (below) no longer holds. `io_capture.h` is retired
+>   from the F5 path (kept in tree for any other in-process capture).
+> - **Syntax highlighting is live.** The compiler's own `Scanner` (already
+>   linked) colors the buffer — keywords cyan, numbers yellow, strings green —
+>   cached per document version (`highlight.h`).
+>
+> Also landed: smart auto-indent on Enter; a modest vi round-out (`w b A I O D`
+> on top of the existing `hjkl 0 $ G gg x dd yy p u Ctrl-r i a o v`); and fixes
+> for the linewise-paste read-only bypass, `want_col` display-column tracking,
+> `Prompt` resize handling, and a `.obs` extension on editor saves.
+>
+> **Deferred, on purpose (not built):** OS clipboard integration; incremental
+> highlight caching (currently a full rescan per edit — fine for REPL-sized
+> buffers); full vi (`d{motion}` composition, `:` ex commands, `/` search,
+> counts, registers); multi-line-string and comment coloring (the scanner
+> reports a multi-line string at its closing quote and never emits comment
+> tokens, so both are left uncolored rather than guessed); char-literal coloring
+> (unreliable token position); find / goto-line; mouse.
+>
+> CI cannot exercise `/e` (no TTY); the gate is a manual terminal matrix.
+
 `obi` today edits by line. `Document`/`Line` hold the buffer and `DoInsertLine`,
 `DoDeleteLine`, `DoReplaceLine`, `DoGotoLine`, `DoList` operate on it through an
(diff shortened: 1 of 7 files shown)


objeck:
**Update — POSIX verified on native Linux (WSL2, Ubuntu 24.04). This closes the "never compiled" gap flagged in the description.**

### 1. The POSIX half compiles and links
`make -f make/Makefile.amd64` in `core/repl` produces a **2.77 MB `obi`**, zero errors. So the `fork`/`execvp`/`pipe`/`O_NONBLOCK` path in `child_run.h` is no longer unbuilt code.

### 2. The subprocess pipeline demonstrably fires
Driving `/e` + F5 under a real pty (`script(1)`), obi left behind **both** artifacts:

```
/tmp/obi-run-14121.obs
/tmp/obi-run-14121.obe
```

That is the whole Part A chain proven end to end: buffer → `doc.Save()` → temp `.obs` → **`obc` child spawned and compiled successfully** → `.obe`. Running that `.obe` with `obr` exits 0.

### 3. Syntax highlighting (B2) confirmed rendering
The raw pty capture contains `ESC[0;36m` around `use`, `class`, `System`, `IO`, `Console` — cyan keywords. Highlighting isn't just compiling, it's painting.

### Still not verified, honestly
My synthetic keystrokes proved unreliable — the temp `.obs` came out with an **empty `Main` body**, i.e. the typed `7->PrintLine();` never reached the buffer, and the editor didn't receive my Ctrl-Q either (the run timed out rather than quitting). So:

- **output-appearing-in-the-pane** and **Esc-cancels-a-runaway** remain untested — the two cases a human must still confirm;
- the F5 escape sequence differs by platform (`CSI 15~` on POSIX, `ESC O T` on Windows), worth knowing for anyone else scripting this.

What the artifacts do establish is that the spawn, argv construction, lib-path plumbing and compile step all work on POSIX. The remaining risk is concentrated in the pump/cancel loop, not in process creation.

PR #648, opened 2026-08-25
perf(vm): cache resolved native entry points, on every platform

objeck:
Every `EXT_LIB_FUNC_CALL` converted the function's name from wide to narrow and
then asked the loader for its address — `GetProcAddress` walks a DLL's export
table, `dlsym` walks the module's symbol hash — **on every call, for a name that
never changes**.

Measured before: **110 ns per call** (69 in the lookup, 39 in the conversion
feeding it). After: a native call goes **2,590 → 2,450 ns, about 5.5%**.

## How

**Thread-local, not shared behind a lock.** A native call is on the hot path of
anything using a native library, and an uncontended mutex would hand back a
useful fraction of what the cache saves. Each thread resolves a given symbol
once.

**A generation counter, not per-entry bookkeeping.** A thread can't purge another
thread's cache, so an unload bumps a counter and every thread drops its cache on
its next lookup. Without it, a library unloaded and a different one loaded at the
same address would be called through a stale pointer. Both unload paths bump it,
as does a failed resolution on Windows — which frees the library before
returning.

One resolver serves both platforms; the `dlerror` dance lives inside it, where
the error can be cleared before the lookup and read straight after.

Misses are cached too — a symbol absent now is absent on the next million calls.

## The profiling, which did not say what I expected

A **no-op** native function costs **2,500 ns** — the same as `glGetError` (2,450)
and `SDL_NumJoysticks` (2,515). The callee's work is irrelevant; it's all VM
overhead. An ordinary Objeck method call is **25 ns**, so a native call is ~100×.

| component | ns | share |
|---|---|---|
| boxing — `Base[1]` + one `IntRef` | 585 | **23%** |
| symbol resolution (this PR) | ~50 | 2% |
| everything else in the VM path | ~1,865 | **75%** |

So this takes 5.5% and leaves the two larger items untouched. **The boxing is
worth naming**: two heap objects per call, and at 1,500 calls a frame that's
3,000 allocations a frame of pure collector pressure. A reusable per-call-site
argument array would take most of it out with none of the risk of the JIT route.

## A measurement I threw away

An in-VM probe timing the inside of `SharedLibraryCall` reported 996 ns "inside
the native function" for `glGetError`, then **1,808 ns for a function whose body
is empty** — slower than real work. The chrono-and-atomic cost per call swamped
what it was measuring. Noting it because those numbers were one step from being
reported as findings; the no-op comparison replaced them and needs no
instrumentation at all.

## Verification

Full Windows regression **188 passed / 3 skipped / 0 failed**. GL suite **603
checks, 0 failed**.

Not covered locally: the POSIX branch compiles here but only CI exercises
`dlsym`. The linux and macOS legs run the full suite.

### DIFF
--- a/core/vm/interpreter.cpp
+++ b/core/vm/interpreter.cpp
@@ -2925,6 +2925,9 @@ void StackInterpreter::SharedLibraryUnload([[maybe_unused]] StackInstr* instr)
 #endif
     }
     (*ext_unload)();
+    // Nothing may hold a resolved pointer into a library that is about to be
+    // unloaded, on this thread or any other.
+    InvalidateSharedLibraryCache();
     // free handle
     FreeLibrary(dll_handle);
   }
@@ -2944,6 +2947,9 @@ void StackInterpreter::SharedLibraryUnload([[maybe_unused]] StackInstr* instr)
     }
     // call function
     (*ext_unload)();
+    // Nothing may hold a resolved pointer into a library that is about to be
+    // unloaded, on this thread or any other.
+    InvalidateSharedLibraryCache();
     // unload lib
     dlclose(dll_handle);
   }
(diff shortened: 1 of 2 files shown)


objeck:
Rebased onto `master` now that #650 has merged (force-pushed; single commit, `45d59ed162` → `25fefb4a1a`).

#650 changed `interpreter.cpp` in the same two functions this PR rewrites, so a clean rebase was not sufficient evidence. Verified after rebasing:

- #650's four `context.alloc_managed_obj = MemoryManager::AllocateObjectNative` bindings all survive, with **zero** stale `AllocateObject` bindings left behind — either side could have been silently dropped.
- This PR's cache (`ResolveSharedLibraryFunction`, `ext_lib_generation`, `InvalidateSharedLibraryCache`) is intact in both the Win32 and POSIX branches.
- The two compose correctly: cached symbol resolution and old-generation native allocation are independent concerns in the same function, and both are present in each platform branch.
- Builds clean (MSVC x64, no new warnings).
- The four GC regression tests from #650 pass against the rebased build, so the symbol cache does not undo the barrier fixes.
- Full regression: **192 passed, 3 skipped, 0 failed.**

One interaction worth noting, which only exists because of the rebase and is correct as it stands: on a failed symbol lookup the cache is invalidated *before* `FreeLibrary`, so nothing retains a function pointer into a library being unloaded.

PR #649, opened 2026-08-25
perf(gl): 4.5x faster frames — hold constant strings and reuse argument buffers

objeck:
> **Stacked on #647.** `.gitattributes` marks `*.obs` binary, so two branches
> touching `sdl_gl.obs` cannot merge in parallel. Retarget to `master` once #647
> lands.

| 500-box frame | | |
|---|---|---|
| Before | 15.02 ms | 67 fps |
| **After** | **3.35 ms** | **299 fps** |

2000 boxes: **52.67 → 13.78 ms**. Medians of three runs.

Two changes, one of which turned out to matter far more than the other.

## 1. A string literal allocates on every evaluation

```
CallFunction("sdl_joystick_nums", a)   1,655 ns
CallFunction(hoisted, a)                 130 ns
```

Same call, same arguments — the only difference is whether the name was written
at the call site. **92% of a native call was allocating and copying a string that
never changes.**

It compounded because the library wrote literals at two layers: each wrapper
named its native function at the call site, and `Scene` named the uniforms at
*its* call sites. Per box that's **seven** evaluations — `"mvp"`, `"model"`,
`"opacity"`, `"tint"`, `"shininess"`, `"specular_strength"`, plus the function
name. At 500 boxes, ~3,500 short-lived strings per frame to name compile-time
constants.

## 2. Argument buffers are reused

`Mesh` and `Shader` build their `Base[]` and boxes once and refill them. The refs
are mutable (`IntRef->Set`, `FloatArrayRef->Set`), so no FFI change was needed.

**Objects passed in stay collectable**: a reused buffer keeps whatever it last
held reachable, so `Shader` would have pinned the last matrix and name it was
given. Both slots are released the moment the call returns. `Mesh`'s buffer boxes
only ints and retains nothing.

**The generational collector sees it**: this turns young→young references into
**old→young**, exactly what a minor GC can't see unaided. `STOR_INT_ARY_ELM` ends
in `WriteBarrier(array - 2)` and field stores are barriered on the instance path
— checked in the source, not assumed. `core_thread_gc_stress` passes.

## How the string cost was found

Disabling `SharedLibraryCall` entirely — returning at the top — moved a native
call from 1,795 → 1,620 ns. So the whole body, including the symbol lookup and
the native call itself, was **175 ns**, and 90% of the cost was spent before the
function was entered. That ruled out everything inside the FFI; hoisting one
string in a throwaway benchmark explained the rest in a single measurement.

Worth recording what this replaces: the plan was to teach the JIT to emit the
call directly — changing the opcode's operand convention across the compiler, the
interpreter and both JIT backends, in the area with the null-frame crash history
— to optimise a 175 ns body.

## This is not just this library

Every Objeck program pays this in any hot loop containing a literal, and nothing
in the source suggests it. The obvious VM fix — interning literals — is **not
safe as things stand**: `String` is mutable, so a program mutating a
literal-derived string would corrupt every other use of the same literal. It'd
need copy-on-write or a compiler escape proof. Raising it rather than reaching
for it.

## Verification

**611 checks, 0 failed.** Full Windows regression **188 passed / 3 skipped / 0
failed**, including `core_thread_gc_stress` and the ODBC and ML native suites.
All eight GL demos compile; the walkthrough and FPS demo run clean; the arena
harness reports `RANGE OK`.

Break-tested: matrix refill removed → the reuse checks plus four existing ones;
mode refill removed → exactly one, the intended one.


objeck:
Closed automatically by GitHub when its base branch `feat/gl-world-culling` was deleted on #647's merge, and GitHub refuses to reopen a PR whose base is gone. Continued as #653 — same branch, same content, rebased onto master.

Recording the mechanic so it does not bite the next stacked PR: deleting the base branch of an open PR closes it irrecoverably. #651 was stacked on this branch, so this branch is deliberately **not** being deleted on merge until #651 is retargeted.

PR #651, opened 2026-08-26
perf(libs): stop allocating the native function's name on every call

objeck:
Rebased onto master (which now carries #650, #647, #648 and #653).

## The cost

An Objeck string literal is materialised into a fresh `String` on **every evaluation** — not once at load, not interned. Every native call in the standard library wrote its entry-point name as a literal at the call site, so every call allocated and copied that name before doing anything else.

Measured on Windows x64 against a native function that does nothing: **1,655 ns** with the name as a literal, **130 ns** with it held in a field. The literal was 92% of the cost of a native call, in code that reads like a constant.

## Scale

**560 distinct native names across 577 call sites** (577 includes 17 sites whose names were already held in hand-written fields before this branch).

| library | names | sites | | library | names | sites |
|---|---|---|---|---|---|---|
| sdl2 | 329 | 337 | | diags | 21 | 21 |
| odbc | 61 | 61 | | ml_core | 14 | 14 |
| opencv | 50 | 52 | | web_server | 13 | 13 |
| sdl_gl | 39 | 40 | | cipher | 9 | 9 |
| onnx | 23 | 29 | | lame | 1 | 1 |

Earlier revisions of this PR said "533", which was a transposition of 593 — and 593 was itself inflated by counting matches inside comment text (see the third commit).

## The commits

1. **The transformation.** Each name held in a static `String` field, filled lazily. Static because the names are class-wide constants and a `function :` cannot read an instance field. The lazy fill races harmlessly between threads: both write an equivalent String, one wins, and nothing can observe a half-built value because the store is a pointer store.
2. **A runtime test for LAME**, which had none — it was compiled by CI and never run, so its single native entry point went unexercised through a change that rewrote its call site. (It also turns out to live in the `Game.SDL2.Aux` bundle, which is part of why nobody had tested it.)
3. **A regeneration fix.** `sdl2.obs` and `sdl_gl.obs` had been transformed by an *earlier* version of the tool that did not skip comments, and were never regenerated after it was fixed. That left 25 permanently-`Nil` static fields and 24 guards spliced into comment text (inert — `sdl2.obl` contained none of those names). The real problem was that the committed files were **not reproducible from their own tool**, which is exactly what makes `--check` worth having. Both regenerated and verified deterministic.
4. **A runtime test for ONNX**, which had none on any platform. `GetProviders` is the one entry point that needs no model file on disk, so it asks the narrow question that was nevertheless unanswered: does the native library load, does a call cross into it, and does a `String[]` come back intact?
5. **The bug that test found** — see below.
6. **Two CI guards** the work needed and did not have.

## What the ONNX test found

It failed on `macos-arm64` only. Every assertion passed, the program printed `All tests passed` and returned 0 — and then the process died with exit 134:

```
libc++abi: terminating due to uncaught exception of type
std::system_error: mutex lock failed: Invalid argument
```

The `Ort::Env` was held in a namespace-scope `std::unique_ptr`, so ONNX Runtime's global teardown ran from a **static destructor**: after `main` returned, while every other image in the process was finalizing too. `OrtEnv::Release` locks a static mutex inside `libonnxruntime`; locking one that has already been destroyed fails with `EINVAL`, and a throw out of a destructor is a `terminate`. linux-x64 survives the same ordering, which is why this only ever showed on macOS — and it showed at all only because the binding finally has a test that runs it.

The Env is now held by a raw pointer, so it has no static destructor and is simply leaked at process exit; `unload_lib` stays the one place it is released, and that runs while the process is alive. `onnx_runtime_test` now runs and passes on macos-arm64 — 194 passed, 0 failed, all five build legs green.

The same read turned up a second defect in the same lifetime: `close_session` reset the **global** Env after deleting **one** session. Every other live session was left pointing at a destroyed Env, and the next `new_session` dereferenced a null one — the QNN and Vitis variants re-create nothing, so they would have taken the null straight through.

## Two CI guards

- **`--check` now runs in CI.** It was written as the durable half of the tool and then never run anywhere. It sits in the Tools job next to the existing `lib_src` guard: needs nothing built, fails in seconds. Without it the whole-tree edit is one regeneration away from reverting — which is not hypothetical, it already happened once in this branch (commit 3).
- **macOS crash reports are collected on failure.** `onnx_runtime_test — runtime error (exit 134)` was the entire report: no signal, no frames. macOS writes a full backtrace for every SIGABRT and SIGSEGV to `DiagnosticReports`, so those now land in the artifact that already uploads on failure. Best effort and non-gating.

## Verification

- `tools/cicd/hoist_native_names.py --check` reports no remaining literals across all libraries — now enforced by CI rather than asserted here.
- Regenerating twice from the same input is byte-identical.
- **All ten committed `core/lib/*.obl` are byte-identical to fresh rebuilds** from their committed sources, built in dependency order with the canonical commands from `core/compiler/update_version.sh`. A stale `.obl` reddens exactly one CI platform and reads as an unrelated test failure, so this was checked rather than assumed.
- The composed `sdl_gl.obs` carries all three of master's culling fix, #653's buffer reuse, and this branch's hoisting — checked by reading, since `.obs` is binary in git and the diff shows nothing.

## Known, and not fixed by this PR

**`web_server` is still run by no test**, and cannot be: its native module is not in this repo. `web_server.obs` resolves it through `Runtime->GetProperty("OBJECK_LIB_WEB_SERVER")`, and none of its thirteen entry points has an implementation anywhere in the tree — they are supplied by the hosting server. Its call sites were rewritten by the same tool as everything else and are covered by `--check`, but nothing here executes them.

The transformation is nonetheless structurally safe: the tool derives each field name from the literal it replaced at that same site, so it cannot change *which* function is called, and every site was verified to satisfy `@fn_X := "X"` with the field declared. The failure mode is also loud rather than silent — a wrong name makes the VM print `Runtime error calling function: X` and exit at the first call.

**`core/lib/sdl/code_gen` generates `sdl2.obs`**, and `sdl_emitter.obs` still emits the literal form, so regenerating it would revert sdl2's 329 hoists. That is now caught by CI rather than only by the tool nobody ran, but the emitter itself is unchanged.

## Review note

`.obs` is binary in `.gitattributes`, so the diff is unreadable. `--check` reporting clean is the machine-readable substitute.


objeck:
## Measured runtime coverage of the hoisted sites

I had been asserting coverage from static reachability and getting it wrong in the optimistic direction, so I measured it instead: the VM was temporarily instrumented to log every native entry point it actually **resolves**, placed on the symbol cache's miss path so each name is recorded once per thread, and the full 193-test suite was run against it.

| library | declared | exercised | | unexercised |
|---|---|---|---|---|
| sdl2 | 329 | 30 | **9%** | 299 |
| odbc | 59 | 20 | 34% | 39 |
| opencv | 50 | **1** | **2%** | 49 |
| sdl_gl | 39 | 39 | **100%** | 0 |
| diags | 21 | 16 | 76% | 5 |
| ml_core | 14 | 7 | 50% | 7 |
| web_server | 13 | 0 | **0%** | 13 |
| onnx | 11 | 0 | **0%** | 11 |
| cipher | 9 | 2 | 22% | 7 |
| lame | 1 | 1 | 100% | 0 |
| **total** | **546** | **116** | **21%** | |

(546 distinct names against 560 declarations — onnx and a few others declare the same name once per class, so 14 declarations are duplicates.)

### What this changes about how to read this PR

**79% of the transformed sites are executed by no test.** "The suite is green" is therefore weak evidence here, and the change's safety rests on its structure rather than on coverage:

* the tool derives each field name from the literal it replaced *at that same site*, so it cannot change which function is called;
* every site was verified to satisfy `@fn_X := "X"` with the field declared;
* the failure mode is loud, not silent — a wrong name makes the VM print `Runtime error calling function: X` and **exit** at the first call, so an unexercised site is untested rather than unsafe.

### Corrections to earlier claims in this PR and its discussion

* `gl_context_test` does **not** broadly cover sdl2 — it covers `sdl_gl` completely (39/39, the only library at 100%) and 9% of sdl2.
* cipher is **2 of 9**, not "well covered by many tests".
* my static odbc estimate of 27/59 was optimistic; measured is **20 of 59**.
* an earlier review finding of "1 of opencv's 52 names" was dismissed here as unverified. It was **correct** — measured 1 of 50.

The instrumentation was temporary, is not part of this PR, and the VM has been restored and rebuilt.

### Follow-up worth filing separately

onnx (13 names) and web_server (11 names) are exercised by nothing on any platform — pre-existing, not introduced here. Those are the two libraries where a genuine bug would be silent.

objeck:
Correction to the last line of the previous comment: it has the two counts swapped. It is **web_server (13 names)** and **onnx (11 names)** — as the table itself shows. Everything else in that comment stands.

Issue #681, opened 2026-08-30
Web server: a program that starts WebServer->Serve and returns from Main segfaults on exit


objeck:
## Symptom

Any Objeck program that starts `Web.HTTP.Server.WebServer->Serve` on a thread and then returns normally from `Main` dies with **SIGSEGV (exit 139)** during teardown, with the server thread parked in `Accept`.

All application output is lost, because stdout is buffered and the process dies before the flush. That makes it look like the program did nothing, when in fact it ran to completion — the crash is strictly after `Main` returns.

## Reproducer

```objeck
use Collection, System.Concurrency, System.IO.Net, Web.HTTP.Server;

class SrvH from HttpRequestHandler {
  New() { Parent(); }
  method : ProcessGet(request : Request, response : Response) ~ Bool {
    response->SetCodeContent(200, "ok");
    return false;
  }
  method : ProcessPost(request : Request, response : Response) ~ Bool {
    return ProcessGet(request, response);
  }
}

class SrvExit {
  function : Main(args : String[]) ~ Nil {
    server := Thread->New() {
      New() { Parent() }
      method : public : Run(arg : System.Base) ~ Nil {
        WebServer->Serve(SrvH->New()->GetClass(), 19991, false);
      }
    };
    server->Execute(Nil);
    Thread->Sleep(700);

    sock := TCPSocket->New("127.0.0.1", 19991);
    sock->WriteString("GET /x HTTP/1.1\r\nHost: 127.0.0.1:19991\r\n\r\n");
    status := sock->ReadLine();     # 'HTTP/1.1 200 OK' -- the request works
    sock->Close();
  }
}
```

The request succeeds. The crash is on exit.

## Scope

**Pre-existing on master.** Confirmed by reverting `net_server.obs` to `HEAD`, rebuilding `net_server.obl`, and rerunning: **3 segfaults out of 3**. Nothing from the `Web.Server` work is involved — this reproducer only touches `Web.HTTP.Server`.

Windows x64. Not yet checked on Linux or macOS.

## Why no test caught it

No in-tree test both starts `WebServer` and exits normally. The server-side tests either drive raw sockets (`core_http_server`) or run a server that is torn down with the process still inside it. The new `web_server_test.obs` hits it directly, and calls `Runtime->Exit(0)` to work around it — otherwise the suite, which scores by exit code, would record a fully passing test as a crash.

## Related: the accepted socket class can be dropped by the linker

[...]

PR #685, opened 2026-08-31
fix(obd): don't free the debuggee's program while a thread is parked in a syscall

objeck:
Follow-on to #683. The same lifetime defect was already sitting in obd, which hosts the debuggee's VM in-process — and there it is **not** Windows-only.

## What obd was doing

`Debugger::ClearProgram` runs after every single run (via `ClearReload`) and:

- deletes the `Loader`, freeing the bytecode buffer and the `StackProgram`;
- calls `MemoryManager::Clear()`, which frees the young region and every old-generation object **and deletes the critical sections the VM takes on the way in** — so a woken thread hits freed heap and destroyed locks, not merely a dangling program;
- called neither `HaltAllExcept` nor `WaitForThreadsToDrain`. Those strings did not appear anywhere under `core/debugger/`. obd was strictly worse off than pre-#683 `vm.cpp`, which at least halted and waited.

And the drain would have lied if it had been called: `WaitForThreadsToDrain` returns true at `intpr_threads.size() <= 1`, i.e. it assumes the caller's own interpreter is registered. `obr` registers its (`vm.cpp:77`); obd never did, so one parked debuggee thread would have read as *"nothing left running"*.

## Reproduced, in two shapes

| Shape | What wakes the parked thread | Windows | Linux |
|---|---|---|---|
| CLI: run, then a client connects to the parked port | an ordinary TCP connection | **6/6** `0xC0000005` | **2/2** SIGSEGV |
| DAP: run to completion, then disconnect | `WSACleanup` at exit | **8/8** `0xC0000005` | 0/2, clean |
| Control: same program, thread exits before `Main` does | — | 0/8 | — |

The program printed its own `PASS` every time — it ran to completion under the debugger and obd died on the way out.

**That split is the headline.** #681 was filed Windows-only because `WSACleanup` was the only thing that ever woke the thread and Linux has no equivalent. obd outlives the run — it goes back to its prompt — so an ordinary client connect wakes the thread just as well, with no `WSACleanup` anywhere near it.

## The fix

- **`debugger.cpp`** — halt and drain in `ClearProgram`; on timeout, `loader->Abandon()` *and* skip `MemoryManager::Clear()`, leaking both rather than pulling them out from under a live thread. Register/deregister the run's interpreter so the drain's own predicate is true (and `RemoveThread` before the delete: obd builds a fresh interpreter per run, so the old pointer would otherwise be handed to the next `HaltAllExcept`).
- **`debugger.cpp`** — no `WSACleanup` on the way out, on any of the three paths. **This is defense-in-depth, not the fix**, and the ablation is why I can say so: with the drain in place, restoring `WSACleanup` does *not* crash (8/8 clean), and the CLI shape proves removing it alone would not have been enough. It buys nothing — Windows reclaims Winsock at exit regardless. The CLI one was unreachable anyway (`Debug()` is a `while(true)` with no break; quit exits through `exit(0)`), and the usage-error one had no matching `WSAStartup` at all.
- **`loader.h`** — `Abandon`'s comment said "used at process exit". obd calls it per run, where the leak is bounded by the debug session.

No iteration-counted sleep timeouts to fix here — there are no sleeps in `core/debugger` at all. The only one on this path was `WaitForThreadsToDrain`'s, already fixed in #683; obd now actually *calls* it, which is what makes that fix matter here. Measured: 2.1s of real drain at exit, not the ~30s it would have been.

## A second, unrelated defect this surfaced

**obd could not debug any multithreaded program.** Compile one with `-debug` — which is how you compile a program you intend to debug — and obd segfaulted on the first instruction the spawned thread executed. `StackInterpreter::debugger` was set by exactly one constructor; spawned VM threads use the default one, which left it indeterminate, and `Execute`'s hot loop calls through it unconditionally in a `_DEBUGGER` build.

It looked like a `-debug` problem rather than a threading one because `ProcessInstruction` reads nothing off `this` until `GetLineNumber() > -1`; without symbols every line is `-1`, so the call returned before dereferencing the garbage. A 2×2 matrix — threads × `-debug` — isolated it in one run, on stock obd.

`_DEBUGGER` is defined only for obd, so `obr`/`obi` never saw this and the VM regression suite could not have caught it. Their preprocessed source is unchanged by this PR. Spawned threads now run with `debugger == nullptr`, i.e. they *run* — breakpoints and stepping still do not reach inside them, which wants the `Debugger` propagated to `AsyncMethodCall`'s interpreter and is a feature, not this fix.

It is a separate commit, and it comes first so history stays bisectable-green: without it the new test's fixture cannot even reach the teardown path under the harness's standard `-debug` compile.

## Test

`programs/regression/obd_teardown_test.py` covers both shapes and reuses `thread_accept_exit_test.obs` rather than adding a near-duplicate fixture — same defect, same program, different host. It asserts the program actually reached the parked-accept state, so a fixture that died on its first syscall cannot produce a green no-op.

It has teeth: against a build with only the teardown commit reverted it reports **5 failures on Windows and 2 on Linux**. Wired into both DAP runners — the `.py` one CI uses and the `.sh` one it does not.

## Verification

- **Windows x64**: DAP 21/21, debugger 25/25, regression 203 passed / 3 skipped / 0 failed.
- **Linux (WSL2, gcc 13.3)**: DAP 21/21.

## Note for the v2026.8.4 release

v2026.8.4 is still a draft and #684 is staging its notes. Those notes cover the VM teardown crash but not these two obd fixes — worth a line each in `CHANGELOG.md` and `readme.json.in` if this lands before the tag. I have not touched #684's files to avoid conflicting with it.

### DIFF
--- a/core/debugger/debugger.cpp
+++ b/core/debugger/debugger.cpp
@@ -109,9 +109,16 @@ static int objeck_main(int argc, const char* argv[])
       std::cerr << "fatal: unhandled exception in DAP adapter: " << e.what() << std::endl;
       return 1;
     }
-#ifdef _WIN32
-    WSACleanup();
-#endif
+    // No WSACleanup() here. The process is exiting, so Windows reclaims every
+    // Winsock resource regardless -- but WSACleanup also unblocks threads still
+    // parked in accept()/recv(), and obd hosts the debuggee's VM in-process: a
+    // program debugged to completion can leave a server thread sitting in
+    // Accept() exactly as it does under obr. DapRun has already released that
+    // run's program image and heap by the time we get here, so waking such a
+    // thread sends it executing against freed memory. That is issue #681, one
+    // layer up: 8 access violations in 8 DAP sessions became 0 in 8 with this
+    // call removed. Same reasoning as win_main.cpp and the getaddrinfo failure
+    // path in arch/win32/win32.cpp.
     return 0;
   }
 
@@ -190,16 +197,16 @@ static int objeck_main(int argc, const char* argv[])
     // go debugger
     Runtime::Debugger debugger(file_name_param, base_path_param, args_param);
     debugger.Debug();
-#ifdef _WIN32
-    WSACleanup();
-#endif
-
+    // No WSACleanup() here either, for the reason given on the DAP path above.
+    // Debug() does not return today -- its command loop exits through the quit
+    // command's exit(0) -- so this was unreachable rather than harmless, and it
+    // would have been the same use-after-free the moment it stopped being.
     return 0;
   }
   else {
-#ifdef _WIN32
-    WSACleanup();
-#endif
+    // Nothing to clean up: this path never reached WSAStartup, so the
+    // WSACleanup that used to stand here was an unmatched call that could only
+    // return WSANOTINITIALISED.
     std::wcerr << usage << std::endl;
     return 1;
   }
@@ -600,6 +607,10 @@ void Runtime::Debugger::ProcessRun() {
     long start = clock();
 #endif
     interpreter = new Runtime::StackInterpreter(cur_program, this);
+    // Register it the way obr does (vm.cpp). ClearProgram's drain asks whether
+    // anything besides this interpreter is still registered, so leaving it out
+    // made one parked debuggee thread read as "nothing left running".
+    Runtime::StackInterpreter::AddThread(interpreter);
     interpreter->Execute(op_stack, stack_pos, 0, cur_program->GetInitializationMethod(), nullptr, false);
 #ifdef _TIMING
     std::wcout << L"# final stack: pos=" << (*stack_pos) << L" #" << std::endl;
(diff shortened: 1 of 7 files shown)


objeck:
CI on the first push was 14/15: the two fixes built and passed on every platform, and the other 20 DAP tests passed on macOS too. The one failure was **this PR's own new test** timing out at 300s on macos-arm64.

Two of its problems needed no macOS to see, being arithmetic:

- **Its worst case exceeded the runner's cap.** 90s waiting for the program's PASS line plus 60s for obd to exit, twice, is 304s before the DAP shape ran at all.
- **It reported nothing when killed** — `TimeoutExpired` carries no partial output, so the log said `timed out after 300s` and nothing about which phase.

The third is a guess I can't confirm from here, so the test no longer depends on the answer. It used to write `r`, wait, then write `q` later. obd's command loop repeats the previous command on an empty read (`Debug()` is a `while(true)`; an empty line becomes `last_cmd`), so a `readline()` that returns rather than blocks on a pipe would re-run the program forever instead of quitting — and macOS resolves `-lreadline` to libedit, whose non-tty behaviour differs from GNU readline's. The existing `.sh` debugger tests never hit this because they pipe every command at once and close stdin.

So the CLI shape now writes `r` and from then on only observes. That's a better assertion rather than a workaround: the crash was on the *wake*, so "obd is still healthy after the wake" is the property, and it needs no second command. Output is drained by a reader thread, so buffering can't wedge it either.

Two further corrections while I was in there:

- `crashed()` no longer means "non-zero". The debuggee sets obd's exit code itself — the fixture calls `Runtime->Exit(1)` when *its own* assertions fail, e.g. the port being busy — so that would have been a false failure. A crash is now a signal on POSIX or an NTSTATUS-range code on Windows.
- Iteration count is bounded by a global 150s budget instead of a fixed count picked to be safe in the worst case. 3 CLI + 2 DAP, skipped and *reported as skipped* if time runs short.

**The teeth improved rather than regressed:** against a build with only the teardown commit reverted it now reports 2–4 failures on 3 runs out of 3, where the previous version needed luck to catch the DAP shape. Clean 3/3 on Windows and 2/2 on Linux against the fixed build, 28s a run; full DAP suite 21/21 on both.

No production code changed in this push — `git diff` against the previous head touches only `obd_teardown_test.py`.

objeck:
macOS diagnosed — the diagnostics from the last push did their job, turning `timed out after 300s` into the actual failure:

```
[FAIL] CLI run 1: program reached the parked-accept state: no PASS within 25.0s;
       output so far: "...loaded binary: '...'\nsource file path: '...'\n\n"
[FAIL] CLI run 1: client woke the parked thread: nothing was listening on 19993
[PASS] DAP run 1/2: obd exited cleanly on disconnect
```

obd printed its banner and **sat there** — the `r` was never processed. Both DAP iterations passed on macOS.

**Cause:** obd's POSIX CLI reads commands through `readline()`, which does not hand back a line from a non-tty on macOS, where `-lreadline` resolves to libedit. A pipe was never the supported shape — `run_debugger_tests.sh` has driven obd's CLI through `expect` (which allocates a pty) on every POSIX platform for as long as it has existed. **The test was the outlier, not obd.** It now uses a pty on POSIX and keeps the pipe on Windows, which has no pty and reads via `std::getline`.

Three further corrections, each measured rather than assumed:

- **Dropped the `> ` prompt wait.** readline writes the prompt to the terminal itself and it never reaches the pty master, so that wait could only ever expire — a probe showed it *never* arriving in 20s, costing ~13s per iteration, which was the whole of a 67s run that should take 23s. It waits for the last banner line instead, which does arrive.
- **Waits for port 19993 to be free before each iteration.** Three iterations start in quick succession against one fixed port, so a previous obd's listener could still be up when the next calls `Listen()` — the program then never reaches the parked-accept state and the iteration burns its budget failing for a reason unrelated to the defect. Observed once, at 65s, against a **fixed** build. A false failure is worse than a missed detection.
- **An all-skipped run now fails** instead of reporting 0 passed / 0 failed, which exit-code scoring would have read as green.

**Verification after the change** — Windows: 13/13 on 4 consecutive runs, full DAP suite 21/21 on 4 more, CLI debugger suite 25/25. Linux: 13/13, suite 21/21 twice. Teeth re-checked against a build with only the teardown commit reverted: **3–4 failures on 3 runs out of 3**. Runtime 23s Linux, ~32s Windows.

**One caveat I have not proved:** a single full-suite run reported 20/1 immediately after a batch of deliberately-crashing teeth binaries, and I did not capture which check it was. Four consecutive clean suite runs followed with no other change. CI does not run crashing builds back to back, so I am treating it as an artifact of that sequence — but I am flagging it rather than calling it resolved.

No production code changed in this push either; the diff against the previous head touches only `obd_teardown_test.py`.

PR #687, opened 2026-09-02
fix(release): three tools that reported success while measuring the wrong thing

objeck:
All three surfaced during the v2026.8.4 release. They share a shape: **a check that passes because it inspects a label, a cached value, or a suppressed error rather than the thing it claims to verify.**

## 1. `update.sh` — the playground served a 2026.6.1 engine for months

The sandbox image is assembled from `core/release/deploy`. That directory is **gitignored**, so `git pull` cannot refresh it, and *nothing in the deploy path ever rebuilt it* — `build-sandbox.sh` only copies out of it. The image was rebuilt faithfully on every deploy, from a tree last built by hand on **June 14**.

`/api/health` read `v2026.8.4` throughout, because that string is a hand-maintained constant in `backend/app/config.py`. Running one line of code through the playground printed `2026.6.1`. Users were getting none of this release's JIT immediate fixes, GC lifetime fixes, `Web.Server` rewrite or `Game.OpenGL`.

**Now**: installs the published `linux-x64` tarball, verifying SHA256 against the published manifest *before* swapping, keeping the previous tree as `.prev` until the new one is proven, and stamping the installed version so repeat deploys skip the download.

Installing the release rather than building on the VPS is deliberate — 1 core, 1 GB RAM, no `cmake`, no `libnghttp2-dev`; an LTO link would likely OOM, and a failed build on the live host is worse than a stale one. It's also the exact artifact users receive, so the playground can't drift from the release again.

Two more fixes in the same file:
- **The pull now recovers instead of aborting.** Files land root-owned and read-only, git can't unlink them, and every pull since has died half-applied — which is how the server silently stopped tracking master.
- **The final check runs code through the sandbox** and asserts the engine version. The health check alone is what let this rot.

## 2. `post_release.sh` — gates that failed like the defects they hunt

**Gate 4** sent its download errors to `/dev/null` and then read the file anyway, so a failed fetch surfaced as:
```
post_release.sh: line 123: /tmp/tmp.XXXX/SHA256SUMS: No such file or directory
```
which reads as a broken script. Its four downloads also shared a single `&&`, so a failure of the first didn't stop the rest.

**Gate 5** reported `could not set release body` for a file that `gh release edit` accepted without complaint when run by hand a minute later — with the reason discarded, there was nothing to act on.

Both now check what they depend on and surface `gh`'s actual error. **Gate 6** gained the engine check above and passes the version through to `update.sh`.

## 3. `watch_release.sh` — a false red on a healthy build

It selected the newest run for the tag without checking the run belonged to the commit the tag *currently* points at. Re-tagging is the prescribed recovery from a failed build, so the broken case is the one that happens under pressure: the tag was re-cut, this ran before GitHub had created the new run, and it matched the **previous** run — cancelled, on the old SHA — printing:

```
RELEASE-BUILD FAILED: completed/cancelled
The tag is already pushed. Fix on master, delete the tag, re-tag.
```

for a build that had just started healthily. Following that advice would have deleted a good tag and restarted a 70-minute build — the exact panic response this script exists to prevent, generated by the script itself.

Now resolves the tag to a commit and accepts only a run on that commit, retrying while the run is created. **Verified against the real data**: with both runs present it selects `33411313720` (`b2e137c529`, success) and excludes `33407768869` (`372fee20d1`, cancelled).

## Verification

The fixed `post_release.sh` run read-only against the published v2026.8.4 passes **all six gates**:

```
[3] All Windows installers for v2026.8.4 are signed.
[4] ok ×10   (regenerated SHA256SUMS vs published bytes)
[5] all 9 named files exist
[6] health ok, version=v2026.8.4
    engine executes 2026.8.4
 ALL POST-RELEASE GATES PASS for v2026.8.4
```

`update.sh`'s tarball path was exercised by hand on the VPS during the release — SHA256 verified, engine confirmed at `2026.8.4` — which is what this change automates.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -637,11 +637,51 @@ jobs:
         run: |
           rm -f ~/.ssh/obj_key
 
-      - name: Skip notice
+      # Fail, do not skip.
+      #
+      # This printed "Skipping deployment" and reported the job as SUCCESS, so
+      # "Deploy API Documentation  completed/success" meant "did nothing" -- and
+      # had meant that for as long as the secrets have been unset. The v2026.8.4
+      # release went out with objeck.org still serving v2026.8.3 docs, and the
+      # 39 Game.OpenGL pages of the release's headline feature returning 404,
+      # while the pipeline reported the deploy green.
+      #
+      # A missing secret is a configuration defect, not a valid state to pass
+      # over quietly: it makes an advertised part of the release silently not
+      # happen. Fail loudly and name what to set. The escape hatch already exists
+      # and is explicit -- the `skip_docs_deploy` workflow_dispatch input gates
+      # this whole job -- so "skip on purpose" is available without "skip by
+      # accident and report success" being the default.
+      - name: Missing objeck.org credentials
         if: env.OBJECK_ORG_SSH_KEY == ''
         run: |
-          echo "⚠️  objeck.org SSH key not configured. Skipping deployment."
-          echo "To enable: Add OBJECK_ORG_SSH_KEY and OBJECK_ORG_USER secrets"
+          echo "::error::objeck.org SSH credentials are not configured - API docs were NOT deployed"
+          echo ""
+          echo "Set both repository secrets:"
+          echo "  OBJECK_ORG_SSH_KEY  - private key authorized on objeck.org"
+          echo "  OBJECK_ORG_USER     - the ssh user"
+          echo ""
+          echo "The docs that should have shipped are in this run's api-docs artifact,"
+          echo "and on master at docs/api.zip. To deploy by hand:"
+          echo "  rsync -avz --delete api/ USER@objeck.org:/var/www/objeck.org/api/v<VERSION>/"
+          echo "  ssh USER@objeck.org 'cd /var/www/objeck.org/api && ln -sfn v<VERSION> latest'"
+          exit 1
+
+      # Prove the deploy actually landed, rather than trusting rsync's exit code.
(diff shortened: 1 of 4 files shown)


objeck:
Added a fourth fix to this PR: **the API-docs deploy skipped and reported success.**

With `OBJECK_ORG_SSH_KEY` unset, the job printed

```
objeck.org SSH key not configured. Skipping deployment.
```

and passed. So `Deploy API Documentation  completed/success` meant *did nothing* — and has meant that for as long as the secrets have been unset, i.e. this deploy has always been manual while the pipeline reported it green.

**v2026.8.4 shipped that way.** `objeck.org/api/latest` still serves v2026.8.3 and the 39 `game.opengl-*` pages — the release's headline feature — return **404**, while every job in the publish run showed success. The docs themselves were fine and published; only the step that puts them on the website did nothing.

Now:
- **Fails** with the secret names and the manual rsync/symlink commands. The deliberate escape hatch already existed and is explicit — the `skip_docs_deploy` `workflow_dispatch` input gates the whole job — so "skip on purpose" stays available without "skip by accident and report success" being the default.
- **Verifies the result**, for the same reason the playground now runs code instead of reading its own health constant: rsync exiting 0 is not the site serving the new pages. It fetches `objeck.org/api/latest` and fails unless the stamp matches the release.

That makes four instances in one release of a step passing while measuring the wrong thing — after the stale `api.zip`, the two `post_release.sh` gates, and `watch_release.sh`'s run selection. Same fix in each case: check the thing itself, and never discard the reason for your own failure.

PR #690, opened 2026-09-03
Bump to v2026.9.0, and close #669 and #659

### LINKED ISSUE #659: diags/compiler: concurrent analysis is safe only because it is serialized behind one global lock; needs per-ParsedProgram factories
Filed to stop this living only in session notes. **This is design debt, not an open crash** — the crash that motivated it was root-caused elsewhere and fixed. Filing it so the reasoning is not lost, and so nobody re-derives the ~5% figure as if it were still live.

## What is actually in place today

`core/lib/diags/diags.cpp` holds a coarse `static std::recursive_mutex s_diag_global_mutex` (line 48), locked in `SafeCallDiag` — which covers all query wrappers — and in the direct entries `diag_parse_file`, `diag_parse_text`, `diag_tree_release`, `diag_signature_help`, `diag_code_rename`. This took the concurrent-analysis crash from ~100% to ~5%.

The residual ~5% was then explained by something else entirely: the **unlocked JIT code-page allocator** fixed in PR #579 (`PageManager::GetPage`, now mutex-guarded on both backends). `diags` JIT-compiles concurrently like any other workload, so it was hitting that race too. After #579, the LSP multithreaded test ran **0 failures in 15**. See #658 for that history.

## What remains

The coarse lock is a serialization, not a fix. The underlying structure is still shared mutable global state:

- `TreeFactory` is a classic singleton — `static TreeFactory* instance` / `static TreeFactory* Instance()` (`core/compiler/tree.h:3313`, `:3331`) — and `TypeFactory` mirrors it.
- Analysis is not idempotent, and `ScopeTable` cursors are shared (see the comment at `diags.cpp:52-59`).

So concurrent analysis is correct today only because it is serialized. Two consequences:

- **Throughput.** The LSP server is genuinely multithreaded (`server.obs:90` spawns a `LintService`/`LspWorker` thread per connection), and every one of them queues behind one global lock.
- **Fragility.** Any new entry point that forgets `SafeCallDiag` re-opens the hole silently.

The complete fix is **per-`ParsedProgram` factories**, i.e. no global singletons in the parse/analyze pipeline. That is a compiler refactor that touches `obc` as well as `diags`, which is why it was deferred rather than attempted.

## Repro / test material

`programs/tests/diag_concurrent_analysis.obs` — 8 workers × `ParseText` + `GetSymbols`. Deliberately **not** gating: it was flaky when added, for the reasons above.

Related: `diags.obs`'s `Release()` is an explicit method rather than a finalizer, so the residual was never release-time `Clear()`.

### LINKED ISSUE #669: Windows loopback: a server that writes then immediately closes can have the response discarded (RST); real interfaces unaffected
On Windows, a socket peer that calls `send()` and then immediately tears the connection down can have the just-sent data **discarded**. The receiver's `recv()` returns `WSAECONNRESET` (10054) with **zero bytes**, even though `send()` accepted every byte and the data had physically arrived.

Linux is unaffected. This is not an Objeck defect — it reproduces in ~130 lines of plain C++ — but Objeck hits it hard because its own client and server share a process in several regression tests.

## Impact

`core_http_server` fails on roughly 80% of local runs; `core_net_buffer` and the two MCP tests fail intermittently with the same signature. Until #668 these were invisible, because those tests printed `FAIL` and still exited 0.

## Reproduction

Self-contained, no Objeck:

```cpp
// server thread:  accept -> recv(4) -> send(N) -> shutdown(SD_SEND) + closesocket
// main thread:    connect -> send(4) -> recv(N)
// N in {1024, 4096, 8192, 16384, 32768}, 20 rounds
```

Result on Windows x64 (loopback):

```
recv ERR=10054 want=4096 got=0
mode=0 failed=18/50  distinct_local_ports=50  failures_on_a_reused_port=0
```

Same bytecode on WSL2 Linux via the Objeck test: **0 failures in 20 runs**. Windows: 8 of 10.

## Mechanism

1. Server `send(N)` returns N — bytes are in the send buffer and transmitted
2. The client has not yet called `recv()`, so its TCP **delays the ACK**. Measured directly: `closesocket()` with `SO_LINGER` blocks **~125ms every time** — the Windows delayed-ACK timer, not network latency
3. The server tears down while that data is **un-acknowledged**
4. Windows aborts a connection carrying un-ACKed outbound data — **RST instead of FIN**
5. The RST **flushes the receiver's buffer**

Step 5 is why the reader sees `got=0` rather than a partial read.

## What does not fix it

Every close-side option was measured, interleaved to control for run-to-run variance (baseline ranges 18%–46% depending on machine load):

| strategy | failures |
|---|---|
| `shutdown(SD_SEND)` + `closesocket` (current) | ~40% |
| `closesocket` only | 74/200 — worse |
| `SO_LINGER(10s)` + shutdown + close | 42–51/100 — no effect |
| `SO_LINGER(1s)`, no shutdown | blocks >1s per close — unusable |
| drain to EOF, 100ms bound | 32/100 — no effect |
| drain to EOF, 400ms bound (above the ACK timer) | 15/50, 30/50 — no effect |
| asymmetric drain (sender only) | 76/200 — no effect |
| **5ms pause before teardown** | **0/50, 0/100 ×3, 0/200** |

Only wall-clock delay works, and only because it lets the *receiving application* reach its `recv()`, which ACKs immediately.

## Ruled out

- **Objeck VM** — plain C++ reproduces it
- **Architecture** — `IPSocket` forks by OS only; both Windows arches compile the same `win32.h`
- **GC** — `--gc-threshold=2g` (minor GCs suppressed), still fails
- **Thread-per-connection** — a single-threaded accept loop fails too
- **Unread data at close** — measured `unread=0 linger=0/0` at every close
- **`SO_RCVTIMEO`** — removing it: 86/200
- **Ephemeral port / TIME_WAIT reuse** — 100 transfers used **100 distinct local ports**, zero reuse, and failures still occurred

## Proposed handling

Not fixable at the `IPSocket::Close` layer. A `Sleep` there would turn every test green while costing latency on every socket close — a race papered over, not closed.

Suggested instead:
1. Treat write-then-immediately-close as unreliable on Windows and document it
2. Make the affected tests tolerant with a bounded retry (the pattern already used in `programs/regression/http_error_body.obs`), rather than quarantining them
3. Consider HTTP keep-alive so the client closes first, which sidesteps the pattern entirely

## Related

- #668 — made these failures visible; they had been exiting 0
- 713b6074e2 — added the `shutdown()` that the table above shows is still the best of the close variants
- Separately fixed: `IPSocket::WriteBytes` dropped the remainder of a short `send()` on both platforms. Real data loss, but **not** the cause of this.

objeck:
Version bump to **2026.9.0**, plus the two open issues.

## Version bump

Generated files from `update_version.ps1` (version.h, code_doc64.cmd, readme.json, five `.rc`), plus the three the generator does not template: the README badge, the playground's version constant, and `readme.json.in`'s feature list. All 32 `.obl` rebuilt.

This bump is **ahead of any feature work** — every commit since v2026.8.4 is release tooling — so the release skill's step-2d re-check of `readme.json.in` matters more than usual. `CHANGELOG.md`, `docs/readme.txt`, `docs/readme.html` and the README Quick Start block are still `update-docs` territory at tag time.

## #669 — Windows loopback: the sender loses what it just wrote

The library server (#672) and the test suite (#680) were already fixed. What was left was the user-facing affordance: `WriteString` then `Close()` still lost the response and nothing in the API said so.

`TCPSocket` and `TCPSecureSocket` gain `CloseGracefully(timeout_ms := 1000)` — drain until the peer hangs up, then close, so the peer owns the teardown. Bounded on both sides.

| server's last call | responses lost |
|---|---|
| `Close()` | **21 of 180** |
| `CloseGracefully()` | **0 of 180** |

30 transfers of a 16KB response per run, six runs. The new `socket_graceful_close_test` fails every run if the one call is swapped back, so it guards rather than asserts a tautology. `thread_accept_exit_test` was flaking at 3 in 30 on exactly this shape; 20 of 20 now.

## #659 — per-ParsedProgram factories

`~ParsedProgram` cleared the *global* `TreeFactory`, deleting every program's nodes and then the factory itself, so one LSP worker releasing its document freed the AST another was walking. The coarse lock could not help — the damage happens inside the call holding it.

Factories now belong to the `ParsedProgram` and bind per thread; the analyzer cache and the lock moved onto the program. Same test binary, 8 workers × 25 parse/analyze/release cycles: **before** crashes or hangs (a 10-run tally could not complete — hung processes ignore the timeout); **after** 20 of 20 clean. `diag_concurrent_analysis` is promoted from `programs/tests` and now gates.

## Incidental: doc comments silently dropping members

Found because `CloseGracefully` documented itself on the TLS twin and vanished on the plain one. `doc_parser.obs` requires a code fence to be the last thing in a doc comment; anything after it makes the parser reject the whole member, with the build still exiting 0.

Missing from the shipped v2026.8.4 reference: `GL->GetLastError()`, `Mesh->GetRadius()`, `Texture2D->Checker()`, `Date->GetTimeString()`, plus the truncated tail of a dozen Game.OpenGL class descriptions. 20 comments fixed, `check_doc_fences.py` added to CI.

Also: `gen_json.cmd`'s sync of `objk_apis.json` to the VS Code client had **never run** — a literal vertical tab where `clients\vscode` should be. The comment above it warns that this exact copy ships stale to users; it was.

## Verification (windows-x64, local)

| suite | result |
|---|---|
| regression | **205 passed, 3 skipped, 0 failed** |
| DAP | **21 / 21** |
| LSP | **45 / 45** |
| `check_doc_lists.py` | 43 libraries agree |
| `check_doc_fences.py` | clean, self-tested both ways |
| `check_test_exit_codes.py` | all 176 asserting tests exit non-zero |
| `docs/api.zip` | 482 pages, `api/` paths, stamped v2026.9.0 |

The 3 skips are the API-key tests. `.obl` verified byte-identical after a full rebuild against the final sources.

Closes #669
Closes #659


objeck:
Reviewed `235a5cb2a0` (#669) and `fd9330b0dc` (#659). Both look right; one docs coherence issue worth a follow-up.

## #669 — graceful close

`CloseGracefully(timeout_ms := 1000)` drains until the peer hangs up, then closes, bounded by the timeout and a 16-read cap. My initial concern was that this looked like the *bounded drain* the docs had already measured as ineffective — but the commit answers it with a measurement:

| server's last call | responses lost |
|---|---|
| `Close()` | 21 of 180 |
| `CloseGracefully()` | **0 of 180** |

That is the right kind of evidence, and `socket_graceful_close_test.obs` makes it repeatable. The bounded design is a defensible trade: it degrades to the old behaviour only against a peer that neither reads nor closes within a second, and in exchange it cannot park a thread.

**One thing to fix, in docs rather than code.** `docs/windows_loopback_sockets.md` still carries, unqualified:

| attempt | result |
|---|---|
| drain to EOF, 100ms bound | no effect |
| drain to EOF, 400ms bound | no effect |

…and then, a few lines below, recommends a drain to EOF with a 1000ms bound. Both are true — the timeout length is the variable — but the document does not say so, so it reads as contradicting itself. Someone hitting this later will not know whether to trust the table or the recommendation. A clause on those rows ("too short for the peer to finish reading and close; see `CloseGracefully`") would settle it.

Not a merge blocker.

## #659 — per-ParsedProgram factories

The approach is clean. `TreeFactory`/`TypeFactory` keep `Instance()`, but the backing pointer is now `static thread_local`, with `Bind()` and an RAII `ScopedProgramFactories` guard — so the thousands of `Instance()` call sites in `tree.h` need no edit and resolve per-thread. That is a much smaller blast radius than threading a factory parameter through the tree.

The coarse `s_diag_global_mutex` is gone in favour of per-state `state->mutex` plus a narrow `s_registry_mutex`, which addresses the throughput half of the issue rather than only the correctness half — every `LspWorker` no longer queues behind one lock.

Also worth noting: `diag_concurrent_analysis.obs` moved from `programs/tests/` to `programs/regression/`, so it gates now instead of being run by hand.

## CI

14 pass, 0 fail, `Build windows-arm64` still running. Merging once it is green.

PR #712, opened 2026-09-06
fix(examples): the lantern's mouse axes want opposite signs

objeck:
`Turn` takes the raw delta so mouse-right turns right; `Look` takes it negated so pushing the mouse forward raises the view.

Both axes originally carried a `* -1.0`, which inverted the **horizontal** as well and made a dark maze genuinely hard to navigate. Removing both then inverted the **vertical** the other way. Neither uniform choice could be right — the two axes want opposite signs, one each.

## The run that settled it exercised the whole game for the first time

```
The dark closed over you.
Again. Four relics, and the lamp is full.
Relic taken -- 3 still out there.
Relic taken -- 2 still out there.
Relic taken -- 1 still out there.
Relic taken -- 0 still out there.
```

Death, restart on `R`, and a full collection — every path in the demo, in one session. Eleven sessions ago not a single relic had ever been picked up.

### DIFF
--- a/programs/examples/gl_lantern.obs
+++ b/programs/examples/gl_lantern.obs
@@ -783,12 +783,16 @@ class Lantern {
 			@window->BeginFrame();
 			delta := @window->GetDelta();
 
-			# No sign flip on either axis. Both carried a * -1.0, which meant
-			# moving the mouse right turned left and pushing it forward looked
-			# down -- inverted on both axes at once, which reads as the controls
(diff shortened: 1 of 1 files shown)


objeck:
Superseded by the branch that carries this same fix alongside the texture and collision work — keeping them in one PR avoids a conflict on the same lines, since .obs files diff as binary and cannot be auto-merged.

PR #736, opened 2026-09-09
[done -- close] CI bisect: F6 variants with the collector trace

objeck:
Throwaway draft to split the ARM64 failures on #735: master plus only the compiler change (EmitBranch). Closed once the answer is in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #737, opened 2026-09-09
[superseded by #735 -- close] CI bisect: the three JIT fixes and F3 without F6

objeck:
The parts of batch 4 that are green on every leg: the three JIT fixes and F3 (loop locals in registers, AMD64). F6 (short-circuit branches in the compiler) stays in #735 until the ARM64 JIT handles its library bytecode; see below. Design for F3 in [`docs/JIT_LOOP_LOCALS_DESIGN.md`](docs/JIT_LOOP_LOCALS_DESIGN.md).

**Three fixes for bugs on master, each with a probe or test that fails there:**

1. **No JIT loop was ever polled for a GC safepoint on AMD64.** A jump's target is the instruction *after* its label, and the poll was keyed on the label's index. A collection on another thread waited for the whole loop. `jit_gc_safepoint.obs`: on master the spin runs all 4e9 iterations before the allocating thread gets its collection; fixed, it stops after ~1e8. (ARM64 already keyed its poll on the label and was not affected; its poll moves to the same place for symmetry.)
2. **Magic division with the dividend in `RDX`** (batch 2) multiplied the constant by itself. The fixture's `JoinAndFloats` probe.
3. **The 16-bit movers had no REX prefix**, so a char element addressed through `R8`–`R15` was read from `RAX`–`RDI`. Dormant until batch 3 put `R8`–`R11` in the Windows pool; the JSON scanner crashed once pinning changed its allocation. The fixture's `Narrow` probe crashes on master.

**F3 — the hottest integer locals of a loop live in registers (AMD64).** Up to three `Int`/`Char` locals per loop in `R13`–`R15`: loaded at the loop header, read and written as registers inside, stored back by a stub on every exit. Object slots are never pinned; callee-saved registers mean every call out of JIT code preserves them by ABI. `Locals` 0.027 s → 0.020 s, `ArraySum` 0.017 s → 0.014 s. `OBJECK_JIT_REPORT=1` names each pinned loop; `OBJECK_JIT_PIN_MAX` / `OBJECK_JIT_PIN_SKIP` turn it off for bisecting.

**Verified:** Windows x64 — fixture byte-identical between interpreter and JIT with the new probes, flag tests 22/22, the regression suite both normally and with `OBJECK_JIT_THRESHOLD=1`; CI — all five build legs green on this exact tree.

**Why F6 is held back:** with the compiler change, the three ARM64 legs fail five tests in the collector's root scan of a JIT frame (Windows ARM64 gets a wrong total in the GC stress test). The tests' own bytecode is unchanged; the difference is in the standard library, which CI rebuilds with the new compiler, and the AMD64 JIT and the interpreter run that bytecode correctly. So the ARM64 JIT mishandles one of F6's new jump shapes; #736 is narrowing which.


objeck:
Closing: this was a CI bisect vehicle for batch 4, which merged as #735. No longer needed.

Issue #746, opened 2026-09-09
JIT + multithreaded GC: core_thread_gc_stress corrupts on every CI leg once it runs compiled (hidden since 2026-06-11 by the runner's substring marker match)


objeck:
## Summary

`programs/regression/core_thread_gc_stress.obs` reports **35–45 list corruptions on every CI leg** the moment it actually runs with the JIT on. It has never done so in CI before: the regression runners matched the `# JIT_DISABLE` marker as a *substring*, and the test's own comment — "must run with JIT ENABLED. Do NOT add a `# JIT_DISABLE` marker" — was the substring. That sentence was added by 5673cf0d71 (2026-06-11), the commit that fixed the stop-the-world GC bugs the test guards, so the fix has only ever been verified in CI with the interpreter.

PR #745 fixes the runners (whole-line marker) and is the run that exposed this:

| leg | result |
|---|---|
| windows-x64 | `FAIL: multithreaded GC stress - 41 corruption(s)` |
| linux-x64 | 45 corruptions |
| linux-arm64 | 44 corruptions |
| macos-arm64 | 35 corruptions |

ci-build run 34399652932 on `fix/jit-disable-marker` (head 77dea108e5, based on f9211c2d19 — *before* batch 6a's entry compilation, so this is plain auto-JIT). The earlier run on 627876c2ec failed the same way.

On a Windows x64 developer box (24 threads) the test **passes** with the JIT on in both `OBJECK_JIT_THRESHOLD=1` and the default mode, twice. CI runners have 2–4 cores; the corruption is timing-dependent.

## What the test does

Twelve worker threads allocate heavily (forcing frequent cross-thread collections) while `Main` repeatedly spawns and joins them. Each worker builds and walks a linked list and counts values that are not what it wrote; any non-zero count is a corruption — an object collected or moved while still live via a root the collector did not see or did not fix up.

## Where to look

- The collector's view of a **JIT frame's registers and working stack**. `MemoryManager::CheckJitRoots` (`core/vm/arch/memory.cpp`) walks a compiled frame's *memory* slots by the method's declarations. A pointer that lives in a pool register or in the write-through local cache at the moment another thread's collection runs — an allocation is a safepoint, and `NEW_OBJ_INST` inside an expression can follow other live pointers held in registers — is invisible to it, and after a moving minor collection the register holds a stale address. This is the shape `jit_gc_timing_suspect` recorded as "young-gen moves objects, possible register/inline-local fixup holes".
[...]

PR #820, opened 2026-09-13
GC: relocate a frame's self only through a genuine forwarding address (hardening; does not fix #816)

### LINKED ISSUE #816: Windows ARM64, every method compiled: intermittent invalid object reaches a cast in Collection.Map->Insert (http_persistence_test)
## What happened

CI run 34781152657 (PR #814, which changes no VM or library code), job **Build windows-arm64**, step **Run regression tests with every method compiled (Windows)** (`OBJECK_JIT_THRESHOLD=1`): 238 passed, 4 skipped, 1 failed.

```
Running: http_persistence_test...
  FAIL (runtime error)
>>> Invalid object cast: '>>> virtual machine: out of memory <<<
Unwinding local stack (000000D1060FC088):
  method: pos=4, name='Collection.Map->Insert(a:K, b:V)'
  method: pos=3, name='Collection.Map->Insert(a:K, b:V)'
  method: pos=2, name='HttpPersistenceTest->ReadResponse(a:System.IO.Net.TCPSocket)'
  method: pos=1, name='HttpPersistenceTest->Main(a:System.String[])'
```

The same commit's master CI passed this step, and linux-arm64 and macos-arm64 passed both modes in the same run. This is the first failure with this signature in the 40 CI runs on 2026-09-13.

## What it means (analysis by the #810 session)

- **Out of memory is a side effect.** The line comes from `JitCompiler::BridgeExceptionExit` (`core/vm/arch/jit/jit_common.cpp:376`), the only place that prints out-of-memory and then unwinds, so the error came from a JIT callback. That callback was `OBJ_INST_CAST` (`jit_common.cpp:780`): `ValidObjectCast` failed. The error message then printed the source object's class name. `GetClass(mem)` returned non-null, but the name behind it was garbage, so writing it threw `std::bad_alloc`, which the bridge reports as out of memory. Memory was not exhausted.
- **Compiled ARM64 code passed a reference that is no longer a valid object** to a cast in the private `Collection.Map->Insert(key, value, node)`. It casts `key` at ip 9/10, 37/38 and 53/54, and `node->GetKey()` at ip 26/27. The likelier bad value is a stored key read back from an existing TreeNode, or `key` reloaded after a call.
- **Not a test race.** `ReadResponse` builds only Strings (`ReadLine`, `SubString`, `ToLower`, `Trim`), and a valid String always casts to `Compare`. The header loop ends on EOF, error or timeout. The failing run took 0.89 s, as a pass does. The map had only two `Insert` frames. The fixed port 18807 plus `Sleep(400)` is worth replacing, as #810 did for `http_error_body`, but cannot produce an invalid object.
- **Not #722** (fixed in #741). That was a wrong `String->Equals` in the handler, not an invalid object in the client thread.
- **x64 doesn't reproduce.** Windows 11 x64: 400/400 runs passed across `OBJECK_JIT_THRESHOLD=1`, default, and `--gc-threshold=64k` / `1m` with `OBJECK_JIT_THRESHOLD=1`.
- **Checked, no explanation found.** The ARM64 callback register save slots (TMP_X0..X5 at SP+144..184) fall inside the collector's 16-word look-back from `jit_mem-1`. Windows ARM64 LLP64 `long` fields are stored by size. The bridge does not park on entry to a cast callback, so the bad reference existed before the cast.
- **Present in shipped releases.** The ARM64 JIT changes most likely to be involved (direct native-entry calls, argument passing, the X12-X15 pool, D8-D15, float park, `self` reload after a park) are already in v2026.9.1 and v2026.9.2.

## Probes (native windows-arm64)

- **P1:** 200 runs with `OBJECK_JIT_THRESHOLD=1`, then 200 with `--jit=off`, to show whether it depends on the JIT and at what rate.
- **P2:** 200 runs with `OBJECK_JIT_THRESHOLD=1 --gc-threshold=64k`, to show whether it depends on GC: a stale reference to a moved object.
- **P3:** a single-thread Map<String,String> insert stress fixture compared with the same fixture plus an allocating thread, to show whether a second thread is needed.
- **P4:** P2 with `OBJECK_GC_TRACE=1`, keeping the first failure's stderr tail.

objeck:
Fixes #816. **Draft until the reproduction loop passes on ARM64 with this build.**

## Why

Compiled ARM64 code crashed under GC pressure when a second thread allocated. On Windows arm64 the crash was a silent access violation or "Invalid object cast" with a garbage class name; on macOS arm64 it was SIGSEGV at address 0x1. It has shipped since the generational collector (v2026.6.3), and CI's windows-arm64 leg has been hitting it in `http_persistence_test` and `jit_concurrent_compile`.

The cause is in `MemoryManager::FixupRoots`. Every frame fixup, three interpreter-frame paths and the JIT-frame path, relocated `self` with a raw read of the object's `MARKED_FLAG` word:

```cpp
size_t fwd = self[MARKED_FLAG];
if(fwd) *mem = fwd;
```

That word holds a forwarding address only after promotion copies the object. Until then it holds the mark bits, and `GC_MARK_BIT` is 1. A `self` that was not forwarded, but whose word was non-zero, was overwritten, with 1 in the mark-bit case. A compiled frame reloads `self` from `frame->mem[0]` after parking for another thread's collection (#746), then passes it as the receiver of its next call.

The macOS fault capture shows exactly that:
- The callee faults on its first instance load with `INSTANCE_MEM = 0x1`.
- `lr` is in the caller's native call site, which executed `ldr x0, [sp,#self]` with the value 1 immediately before `blr`.
- The other thread was inside `CollectAllMemory`.

## What

- `MemoryManager::FixupSelf(mem, method)` relocates `self` only to an address `ForwardedAddr` accepts: aligned, inside the nursery, and a member of the old generation. That is the check every other slot fixup (`FixupSlot`) already used.
- All four raw reads are replaced. No other code reads `MARKED_FLAG` as a forwarding address.
- Under `OBJECK_GC_TRACE`, it prints `[gc] self not forwarded: method=... self=... header=...` when a young `self`'s `MARKED_FLAG` word is non-zero but is not a forwarding address. This shows whether a `self` can reach the fixup unforwarded, which would be a second bug upstream.

## Reproduction and controls (before this fix)

Fixture `programs/tests/map_insert_stress_816_mt.obs` (branch `probe/816-p3`), command `OBJECK_JIT_THRESHOLD=1 obr --gc-threshold=64k map_insert_stress_816_mt.obe 20000`:

| platform | P3b + gc64k | default GC | single thread | `--jit=off` + gc64k |
|---|---|---|---|---|
| Windows arm64 (CI) | 3/100 fail | 0/100 | 0/100 | — |
| macOS arm64 (M4 Max) | 6/100 fail | 0/50 | 0/50 | 0/50 |
| Windows x64 (7950X3D) | 0/100 | — | — | — |

## Testing plan

- [ ] CI: all five legs build and pass regression, debugger, DAP and VM flags.
- [ ] macOS arm64: this build, the P3b + gc64k loop, 200 runs, 0 failures. With `OBJECK_GC_TRACE=1`, count any `self not forwarded` lines.
- [ ] Windows arm64 and Linux arm64: the same loop against this PR's CI build artifacts.


objeck:
**Correction: this does NOT fix #816.** On macOS arm64 the same loop with this build (20f179b58e, and the `self not forwarded` diagnostic ungated locally) faulted on run 15 of 15. The fingerprint was identical: FAR 0x1, x0 = 1, pc - lr = 0x24bc, the other thread parked in `CollectAllMemory`. The diagnostic never fired in any run.

The raw `self[MARKED_FLAG]` read this replaces was unsafe: `ForwardedAddr` is the correct check and every other slot already used it. But it is not what writes the 1. Keeping this as a draft while the real cause is found. It will either be re-scoped as hardening with a regression test, or folded into the actual fix.

objeck:
Superseded: this change landed on master in refined form as 4542ec7e41 (`fix(gc): relocate a frame's self only through a genuine forwarding address (#820)`), which also requires a genuine young object start before reporting an unforwarded self. The v2026.10.0 CHANGELOG lists it. Closing the draft so it isn't merged over the refined version.

PR #821, opened 2026-09-14
GC: never mark a nursery address that is not an object start (#816)

### LINKED ISSUE #816: Windows ARM64, every method compiled: intermittent invalid object reaches a cast in Collection.Map->Insert (http_persistence_test)

objeck:
Fixes #816. **Draft until the reproduction loop passes on ARM64 with this build.**

## Why

Compiled ARM64 code crashed under frequent collections when a second thread allocated:
- Windows arm64: a silent access violation, or "Invalid object cast" with a garbage class name
- macOS arm64: SIGSEGV at address 0x1
- Linux arm64: segfault

CI's windows-arm64 leg kept hitting it in `http_persistence_test` and `jit_concurrent_compile`. It has been in shipped releases.

The bad value is a **heap field, and the collector writes it.** Evidence:
- **Linux core, walked with gdb:** a young-born `TreeNode` `[key, value, left 0, right 1, level 1]` sat in the old generation. `Map:Find` read `node = node.@right` = 1 and called `GetKey` on it.
- **macOS tripwire build:** caught `TreeNode.@right = 1` at the start of a collection, before any crash.

The mechanism is in `MemoryManager::CheckObject`. It marked whatever a conservative scan handed it: operand stacks, JIT temp windows, monitor stacks. When `GetClassMapping` found no object class, it called `MarkMemory` **before** checking that the memory was an array, and `IsAllocated` accepts any 8-byte-aligned address in the used nursery. A stale or interior nursery pointer therefore had `GC_MARK_BIT` ORed into the word below it. The address of a young `TreeNode`'s `@level` turns its Nil `@right` into 1, and promotion copies the damage into the old generation.

ARM64's conservatively scanned JIT temp window overlaps scratch slots that hold addresses into the shared nursery, so the failure needs:
- compiled code
- an allocating second thread
- frequent collections

## What

`CheckObject`:
- never marks a **nursery address with no class mapping**. The nursery holds objects only; `AllocateArray` always allocates in the old generation.
- requires a **nursery address with a class** to be a real object start. `MemoryManager::IsYoungObjectStart` checks that the allocation size word both young allocation paths write matches the class.
- applies the minor-collection old-object shortcut only to addresses **outside** the nursery, because `IsOldGen` reads the word below the address.

## Reproduction (before the fix)

`OBJECK_JIT_THRESHOLD=1 obr --gc-threshold=64k map_insert_stress_816_mt.obe 20000` (fixture on branch `probe/816-p3`):

| platform | failures |
|---|---|
| Linux arm64 (CI) | 24/150 |
| macOS arm64 (M4 Max) | 6/100 |
| Windows arm64 (CI) | 3/100 |
| Windows x64 (7950X3D) | 0/100 |

Controls, all clean:

| control | result |
|---|---|
| `--jit=off` with gc64k (macOS) | 0/50 |
| default GC (Linux, macOS) | 0/50 |
| single thread (Linux, macOS) | 0/50 |

## Testing plan

- [ ] CI: all five legs pass regression (both passes), debugger, DAP and VM flags.
- [ ] Linux arm64: this build, the same loop, 150 runs, 0 failures.
- [ ] macOS arm64: this build, the same loop, 200 runs, 0 failures.
- [ ] Windows arm64: the same loop against this PR's CI artifact.

#820's `FixupSelf` hardening is independent and can be folded in or closed separately.


objeck:
**Validation progress.**

- **CI:** green on all five legs, 16/16 checks including CodeQL and the perf gate. Regression, both passes, had 0 failures everywhere, as did debugger 45, DAP 24 and VM flags.
- **macOS arm64 (M4 Max), tripwire build:** 5bf8bb286b plus checks that abort at the first collection that finds a small integer or -1 in an object field, running `OBJECK_JIT_THRESHOLD=1 obr --gc-threshold=64k map_insert_stress_816_mt.obe 20000`:

  | build | runs | result |
  |---|---|---|
  | **#821** | 100 | **100 PASS, 0 trips, 0 hangs** |
  | master 0048a4745e, same tripwires, same load | 13 | 4 trips (`Collection.TreeNode [key, value, 0, 0x1, 0x1]`, field 3), 2 hangs at 120 s, 7 PASS |

- **macOS arm64, plain #821 build:** 101/101 so far of 200.
- **Still running:**
  - Linux arm64, 150 runs built from source (was 24/150)
  - Windows arm64 and Linux arm64 against this PR's CI artifacts, 150 each

The hangs on master matched the one hang in the original P3b baseline, and neither fixed loop has hung, so the corruption also accounted for those hangs.

objeck:
**macOS arm64 validation complete.** The plain #821 build (5bf8bb286b, clean worktree, ad-hoc signed, minos 13.3) passed **200/200**: 0 failures, 0 hangs, no stderr output, no crash reports.

Runs 101-200 ran alone, under the same conditions as the pre-fix baseline, which failed 6/100. At that rate, 200 clean runs happen by chance about 0.0004% of the time.

Together with the tripwire build (100/100, 0 trips) and the unfixed control under the same load (6 of 13 failed: 4 `TreeNode.@right = 1` trips and 2 hangs), macOS arm64 is done. Still running: Linux arm64 from source (150 runs) and the Windows/Linux arm64 runs against this PR's CI artifacts.

objeck:
**Linux arm64 validation complete.** `obr` was built from source at 5bf8bb286b on ubuntu-24.04-arm, and the P3b loop (`OBJECK_JIT_THRESHOLD=1 --gc-threshold=64k`) ran **150 times with 0 failures** (run 34792406171, 4261 s).

The same loop without the fix, on the same runner type:
- master's CI build: 24/150 failed
- a source build of master (bisect baseline): 10/60 failed

Still running: Windows arm64 (and Linux arm64 again) against this PR's CI artifacts.

objeck:
**Validation complete on every ARM64 platform.** P3b loop (`OBJECK_JIT_THRESHOLD=1 obr --gc-threshold=64k map_insert_stress_816_mt.obe 20000`) against this PR's own CI build (run 34793397290):

| platform | runs | failures | before the fix |
|---|---|---|---|
| Windows arm64 (windows-11-arm) | 150 | **0** | 3/100 |
| Linux arm64 (ubuntu-24.04-arm) | 150 | **0** | 24/150 |

Also:
- Linux arm64 built from source: 0/150
- macOS arm64 plain: 200/200
- macOS arm64 tripwire build: 100/100 with 0 trips (master under the same load: 6 of 13 failed)
- CI on 9a5a3bb893: green, 16/16

Merging for v2026.9.3, as the maintainer approved.

Issue #861, opened 2026-09-15
core_thread_gc_stress segfaults on Linux x64 with a small nursery (OBJECK_NURSERY=256k)


objeck:
Found by the stress probe workflow (run 35030415215, branch probe/stress-2026-09-15) on the master ad2c610bb0 CI binaries.

- **Linux x64:** `core_thread_gc_stress` with `OBJECK_NURSERY=256k` exited with SIGSEGV (-11) in 1 of 3 runs, with no output on stdout or stderr.
- **The same fixture passed:**
  - on Linux x64: 30/30 default, 30/30 with every method compiled, and 3/3 with the heap verifier plus a 256k nursery;
  - on Linux arm64, macOS arm64, Windows x64 and Windows arm64, in every mode.
- **Every other fixture and mode passed on all five platforms:** 6170 of 6171 runs, most modes at 200 runs.

The fixture runs 12 threads with long-lived lists under heavy allocation, and a 256k nursery makes minor collections very frequent, so a multithreaded stop-the-world or minor-GC race is the leading suspect. It needs reproduction under WSL Linux x64 with a core dump.

The small-nursery modes of this fixture ran only 3 times per mode, so coverage of this combination is thin.

objeck:
**Cross-platform confirmation: macOS arm64 reproduces this and the fix holds there.**

20,000 runs of `gc_mt_small_nursery_stress.obs` at `--nursery=256k` on an Apple M4 Max (14 cores), 6 concurrent copies giving ~48 fixture threads — the same oversubscription that substitutes for `taskset`, which macOS does not offer. Same `lang.obl` on both arms, so only the VM differs.

| arm | jit=off | jit=1 | total |
|---|---|---|---|
| pre-fix | 4 / 5000 | 7 / 5000 | **11 / 10000 (0.11%)** |
| post-fix | 0 / 5000 | 0 / 5000 | **0 / 10000** |

95% upper bound on the post-fix rate: 0.03%. Under equal rates, all 11 events landing in one arm is `2 * 0.5^11 < 0.001` — stated this way because it assumes nothing about the underlying rate, which is the uncertain quantity. Independently, `P(0 in 10000 | 0.11%) = 1.7e-05`.

[...]

Issue #854, opened 2026-09-15
Method call on a string literal crashes when an argument calls a class function


objeck:
Found by the v2026.10.0 EMI fuzzer mode (hardening/fuzz-emi-mode). It crashes in every configuration (-opt s0 and s3, --jit=off and --jit=1) with an access violation or a Nil dereference, and reproduces on master after #853.

```
class RE {
  function : Main(args : String[]) ~ Nil {
    r := "abcdef"->SubString(3, (3 + G->Zero()));
    r->ErrorLine();
  }
}
class G { function : Zero() ~ Int { return 0; } }
```

`"abcdef"->Get(G->Zero())` gives '>>> Attempting to dereference a 'Nil' memory instance <<<'. Controls that pass:
- hoisting the call into a local;
- a variable receiver;
- a String-returning call in the argument.

s0 bytecode: NEW_CHAR_ARY / TRAP_RTRN / NEW_OBJ_INST System.String / MTHD_CALL String:New, then LOAD_INT_LIT 3, SWAP_INT, LOAD_INST_MEM, MTHD_CALL G:Zero, LOAD_INT_LIT 3, ADD_INT, MTHD_CALL String:SubString. The literal receiver sits on the stack while the nested call is emitted, so the SWAP_INT around the nested call is the suspect (compiler emitter, not the optimizer or the JIT).

Affected existing test shapes, found by EMI mutation: fix534_substring_crash, jit_string_ops, string_find_ops.

objeck:
Fixed by #868 (integration batch 8), with regression test programs/regression/string_literal_receiver_nested_call.obs; the v2026.10.0 CHANGELOG records it. The conditional-receiver case it mentioned was a separate defect, #863, fixed in #872.

PR #868, opened 2026-09-16
hardening: integration batch 8 -- the #861 GC fix, and two test-output artifacts

objeck:
Sixth integration batch of the v2026.10.0 hardening work: 24 commits, 43 files,
+4109/-350. It is `hardening/integration-7` plus the #861 GC fix, which was held
out until its independent review finished so that any failure here points at one
change rather than two.

## Bug fixes

- **A collection could crash the VM while another thread was exiting** ([#861](https://github.com/objeck/objeck-lang/issues/861)): a thread whose entry method had returned left its monitor registered with a null current frame after it had already left the stop-the-world count. A collection started by another thread inside that window dereferenced the null frame on a mark thread, so the process died with SIGSEGV and no output — the exact CI signature. It needs a small nursery and more runnable threads than CPUs, which is why a 2-core runner found it and developer machines did not.
- **`each` over an array captured by a lambda would not compile**: a captured variable resolves to a closure-copy entry that is deliberately neither local nor static, and a lambda is compiled as a static method, so the static check rejected it. This is the third site of that exemption; an audit of the remaining callers is queued.
- **A method call with a receiver already on the stack could use the wrong slot as `self`** ([#854](https://github.com/objeck/objeck-lang/issues/854)): found by the new EMI mutation mode.
- **The `-opt s3` inliner shifted a callee's local slots** when the callee contained `and`/`or` and the caller did not, which produced a real use-after-free under the collector.

## Infrastructure

- **EMI mutation mode** for the fuzzer: semantics-preserving mutations (dead-code insertion, unexecuted-block deletion, runtime identities) whose output must not change. This is what found #854.
- **Stress-probe workflow** and loop driver.
- **The nightly is signal-only**, and its verifier opt-outs are scoped to what actually failed rather than blanket-excluded.
- **`binarytrees` runs at the depth `docs/performance.md` publishes.** It had been running at 21 against a documented 17 — 16x the work — which made a benchmarking session estimate hours. `tools/cicd/check_benchmark_inputs.py` keeps the two in step.

## Validation

All on Windows x64 against a tree built from these sources, with every `.obl`
regenerated by the fixed compiler:

| Leg | Result |
|---|---|
| Regression, default JIT | 286 passed, 4 skipped, 0 failed |
| Regression, every method compiled | 286 passed, 4 skipped, 0 failed |
| VM flag tests | 203 passed, 0 failed |
| Differential (s0/s3 x interpreter/default/jit1) | 267 compared, 1 divergent -- characterised below |
| Lints (7 checks) | all pass |
| Unit tests (cicd, regression, fuzz) | all pass |
| Debugger, DAP, LSP | 0 failed |
| Heap verifier, 256k nursery | 0 violations |
| Heap verifier, default nursery | 0 violations; 1 test failed on a socket read -- characterised below |
| Fuzzer | 2000 / 2000 clean |
| EMI over every eligible test | 979 / 980 variants -- the same test as the differential |

## The two red legs were test output and machine load, not defects

Both are fixed in this branch, and neither touches the VM or the compiler.

**`native_gc_leak_test` (EMI).** It prints its peak old-gen bytes to stderr
with a comment saying they vary by JIT mode and that the differential compares
stdout — then interpolated the same peak into a check label, and the helper
prints `PASS: <label>` on success. The diagnostic was moved and the assertion
was not. The interpreter group agreed, the JIT group agreed, and they differed
from each other, both PASSING. The label loses the numbers; the stderr line
keeps them. The rest of the suite was swept rather than assumed isolated:
`runtime_gc_stats` interpolates a mode-dependent occupancy into two labels too,
but its helper prints only on failure, so it cannot diverge on a passing run
and is correctly left alone.

**`mcp_debug_test` (differential).** It enables the MCP server's debug tracing,
so the server library traces every request to stdout, and its own `Request()`
retries a POST after #669's loopback teardown race drops a response — a retried
request is handled and traced twice. Measured across all four compared
configurations: 11 of 12 runs byte-identical, and one run at `s3/jit1` of 205
lines instead of 124, the difference being exactly that extra block, PASS in
both. The odd run was a JIT configuration in one round and not in the next two,
so the difference follows the retry, not the configuration. Marked
`NONDETERMINISTIC_OUTPUT` — exit status is still compared, the PASS/FAIL lines
are still checked by the regression runners, and the non-debug twin
`mcp_server_test` keeps full stdout comparison.

**`core_net_buffer` (heap verifier, default nursery).** Not a GC fault: the
verifier logged **zero** violations, and the failure was `read=-1`, a socket
error. It passed in both earlier regression passes and ran 5/5 clean under the
verifier once the machine was quiet. Four agents were building concurrently
during that leg. `run_differential.py`'s `DIFF_SERIAL` keeps a test away from
other tests in its own pool, but cannot know about other processes on the box.

## What was validated

The suite above ran at `1054dbeb54`. The two commits merged after it are a
check label and a test marker — no code — and the marker only changes what
`run_differential.py` compares, which CI does not run. The artefacts CI builds
are therefore identical to what was validated.


objeck:
Performance check before merge (Windows x64 native, 7950X3D, quiet box): this head (555f321a2d) vs master (ad2c610bb0), all 18 benchmarks from docs/performance.md and programs/tests/perf, interleaved ABAB, adaptive run counts (15-101 per build), each tree compiled with its own obc and its own committed .obl, benchmark sources identical between trees. Every benchmark within ±1.0% of master, none significant (smallest p = 0.28); the >5%-slower-at-p<0.01 gate passes. Peak RSS unchanged (within 0.07%). 2086 runs, 0 failures. The CI perf gate is report-only against an approximate baseline, so this is the measurement the merge rests on.

Issue #841, opened 2026-09-15
Nursery default from a size sweep; Windows commit on growth (G3)


objeck:
Plan: docs/PLAN_2026_10_0_HARDENING.md (revision 2).

Sweep 16/32/64/128 MB on Docker Linux x64, Windows x64 and macOS arm64 (peak RSS, time, GC counts, pauses) and pick the default from data; reserve 128 MB and commit on growth on Windows (every program commits ~132 MB at startup today); keep OBJECK_NURSERY=128m as a kill switch.

objeck:
## Sweep complete on all three platforms — decision: keep 128m for now

Same method on every leg: tree `ad2c610bb0`, `perf-results/objeck_bench.py nursery`, 4 sizes x 3 benchmarks x 15 runs after 2 warm-ups, 0 failures. Raw data and machine state are in `perf-results/baseline-2026.10.0/` (PR #869).

Change vs 128m (time / peak RSS):

| | macOS arm64 | Linux x64 (Docker) | Windows x64 native |
|---|---|---|---|
| binarytrees 16m | +35% / −49% | +81% / −54% | +163% / −56% |
| binarytrees 32m | +20% / −48% | +35% / −45% | +70% / −49% |
| **binarytrees 64m** | **+6.5% / −26%** | **+10.5% / −31%** | **+25.8% / −28%** |
| bench_gc_churn 64m | −0.7% / −48% | −9.2% / −47% | −2.4% / −47% |
| bench_gc_large_heap 16m / 32m | +318% / +319% | +279% / +355% | +991% / +1040% |
| bench_gc_large_heap 64m | +0.9% | +2.2% | +2.3% |

- **16m and 32m are ruled out everywhere.** The live set stops fitting, so 28 MB is promoted (zero at 64m/128m), and large_heap is several times slower *and* uses more memory.
- **64m is only "never bad" on macOS and Linux.** On Windows it costs +25.8% on binarytrees. The cost follows promoted bytes, not collection count: churn does 25 extra minor GCs, promotes nothing, and pays nothing. On the same 7950X3D, Windows promotion is ~3.5x slower than Linux. Filed as #871.

**Decision (maintainer):** keep the 128m default. Fix #871, then re-measure 64m. This issue stays open for that re-measurement and for its second half, Windows commit-on-growth: `VirtualAlloc` still commits the full 128 MB region at startup regardless of `--nursery`.

objeck:
[...]

Issue #840, opened 2026-09-15
Objects are allocated at twice their field size (G2)


objeck:
Plan: docs/PLAN_2026_10_0_HARDENING.md (revision 2).

`alloc_size = inst_size * 2 + header` (memory.cpp:467, memory.h:257, jit_amd_lp64.cpp:1426) although inst_size is already bytes; stale since 5c5d7b2a8e dropped 32-bit support. Guard-word experiment first, then exact sizing if clean.
Branch: hardening/objsize

objeck:
Fixed on master (ad2c610bb0), merged through #852 and #853, and verified by reading the fixed code rather than the commit messages: each has its own regression test, and CI passes on all five platforms in both JIT modes (windows-x64/arm64, linux-x64/arm64, macos-arm64).

PR #869, opened 2026-09-16
perf: macOS arm64 baselines, the #841 nursery sweep, and the #861 arm64 data point

objeck:
Measurement data only — **nothing outside `perf-results/`**, no product code, no compiler or VM changes, no `.obl`, no `TESTS.md`.

```
git diff --name-only ad2c610bb0..HEAD | grep -v '^perf-results/'   ->   empty
```

### Why a gitignored path is being committed to

`perf-results/` is in `.gitignore`, so these files are added with `-f`. That is deliberate and consistent, not a workaround:

- **~30 files under `perf-results/` are already tracked the same way** — `run_benchmarks.sh`, `check_perf_gate.py`, `gen_charts.py`, `perf_baseline.json` among them. The ignore exists to keep *ad-hoc local run output* out of the tree, not the committed harness or its published results.
- **Plan item 0.3 asks for exactly this**: "Baselines, raw samples committed to `perf-results/baseline-2026.10.0/`", with "baselines committed" as its exit criterion.

A reviewer's first question about this PR is likely to be why the ignore is being bypassed, so it is answered here rather than left to be rediscovered.

Measured on an idle Apple M4 Max (14 cores, macOS 26.6.2 arm64, on AC, `caffeinate` held), master against v2026.9.4, ABAB interleaved.

## 1. The DeadStore performance gate PASSES

No benchmark is >5% slower at p<0.01. Master is **faster** on three, all highly significant: `binarytrees` −2.92%, `bench_gc_large_heap` −2.30%, `bench_gc_churn` −2.02%. `bench_array_intensive` is +3.11% — real, but under the threshold.

Peak RSS on `bench_gc_large_heap` is **38.6 MB against 52.6 MB, −26.6%**, which lands exactly on #840's `ObjectBlockSize` fix removing the 2× allocation. An independent measurement confirming a specific known fix is the strongest single number here.

**One result was withdrawn rather than reported.** `bench_matrix_multiply` first read +8.95% at p=0.042 — over the size threshold, under the significance one. Re-measured at n=60 it was −1.58%, master *faster*, p=0.222. Master's own median moved between samples, so the median is simply unstable at n=15 on a 0.2 s workload. Had it cleared p<0.01 by chance it would have entered the record as a 9% regression.

That produced the methodology now in the driver: runs scaled so each benchmark gets ≥10 s of wall time per build, floor 15. 15 runs is right for the CLBG set and under-powered for sub-second micro benchmarks.

## 2. #841 nursery sweep — recommends 64m, pending two more platforms

16m and 32m are **disqualified by a mechanism**, not by a preference: on `bench_gc_large_heap` they are 4.2× slower *and* use more peak RSS (90.8 / 112.5 MB against 38.5 MB), because the live set stops fitting the nursery and survivors get promoted — 28 MB promoted at 16m/32m against **zero** at 64m. A default that quadruples the runtime of a large-live-set program is not shippable whatever it does for churn.

64m is the only size that is never bad: +6.5% on `binarytrees` against 128m, within noise elsewhere, while cutting peak RSS 26–48% and halving churn's p95 pause.

**This does not set the default yet.** One machine, one architecture — and the size that fits a live set is workload- and architecture-dependent by definition. Windows x64 and x64 Docker sweeps are queued; #841 stays open until all three agree.

## 3. #861 reproduced on macOS arm64, and the fix holds

20,000 runs at `--nursery=256k`, 6 concurrent copies (~48 fixture threads on 14 CPUs), same `lang.obl` on both arms so only the VM differs:

| arm | jit=off | jit=1 | total |
|---|---|---|---|
| pre-fix | 4 / 5000 | 7 / 5000 | **11 / 10000 (0.11%)** |
| post-fix | 0 / 5000 | 0 / 5000 | **0 / 10000** |

Under equal rates, all 11 events landing in one arm is p < 0.001, assuming nothing about the underlying rate. All 11 crashes are one signature: `EXC_BAD_ACCESS` at `0x20` in `CheckPdaRoots` — on LP64 that offset is `StackFrame::jit_mem`, read through a null `cur_frame`, which is the line the fix guards.

This evidences the exiting-thread window **only**. It does not demonstrate the `RegisterMutator` ordering race or the `interpreter.h` null-inits; those remain hardening justified by reasoning.

## 4. `jit_closure_call_in_creating_frame` passes on ARM64

All 12 modes, with non-vacuity confirmed by GC counters.

## Driver

The driver that produced these numbers is **not** committed here. It is landing separately as `perf-results/objeck_bench.py` with three corrections this version lacks — `stress` mode refusing to run without CPU affinity instead of emitting a powerless clean number, peak RSS on all three platforms with the instrument recorded per value, and `.exe` handling without which it cannot find `obc`/`obr` on Windows at all. Keeping both would leave the uncorrected copy in the tree under a second name.

The macOS numbers are unaffected: the macOS RSS path is the one that worked, and none of the four items used `stress` mode. The README records which driver produced them and how it differed.


objeck:
Added the Linux x64 (Docker) and Windows x64 legs of the #841 nursery sweep in e257c2b4b1. This changes the conclusion of section 2: 64m costs +25.8% on binarytrees on Windows (vs +6.5% macOS, +10.5% Linux), because promotion is ~3.5x slower on Windows on the same hardware (#871). The README now says the default stays at 128m until #871 is fixed. The commit touches only perf-results/ (4 data files + README).

PR #870, opened 2026-09-16
perf: macOS arm64 baselines, nursery sweep and the #861 arm64 data point

objeck:
Phase 0.3 baselines for v2026.10.0 from the macOS arm64 box (Apple M4 Max, 14 cores, macOS 26.6.2), plus three other measurements the plan and the nightly asked for. **Data and the measurement driver only — no product code is touched.**

Both builds came from their own clean worktrees with their own dependencies: master `ad2c610bb0` and release `v2026.9.4`. Each ran its own libraries, verified by checksum.

## 1. Perf baselines — GATE PASS

18 benchmarks, ABAB-interleaved between builds, 2 warm-ups discarded, 15 runs each, 2136 s. Nothing is >5% slower on master at p<0.01.

| benchmark | master | v2026.9.4 | delta | p |
|---|---|---|---|---|
| binarytrees | 2.0160s | 2.0766s | **-2.92%** | 3e-06 |
| bench_gc_large_heap | 0.0261s | 0.0268s | **-2.30%** | 3e-06 |
| bench_gc_churn | 0.1699s | 0.1734s | **-2.02%** | 4.8e-05 |
| bench_array_intensive | 0.0845s | 0.0819s | +3.11% | 3e-06 |
| bench_matrix_multiply | 0.1979s | 0.2011s | -1.58% | 0.22 (n=60) |

Negative = master faster. Everything else is within ±1%. The three where master wins are all GC-adjacent and all highly significant.

**Peak RSS** (compared on `/usr/bin/time -l`, since v2026.9.4 predates the VM's own counter): `bench_gc_large_heap` **-26.6%** (38.6 vs 52.6 MB), which lands on #840's `ObjectBlockSize` fix removing the 2x allocation; `binarytrees` -3.2%; and a constant **+320 KB** floor on every small benchmark that does not grow with the workload — the `obr` binary is only +67 KB bigger, so the rest is the verifier and stats infrastructure master compiles in.

### A withdrawn result, deliberately in the record

`bench_matrix_multiply` read **+8.95% (p=0.042)** at n=15 — over the gate's size threshold, under its confidence threshold. At n=60 the same pair reads **-1.58% (p=0.22)**, and master's own median moved 0.2199 → 0.1979 between samples. The instability was in the sampling, not in either build.

The driver now sets `n = clamp(ceil(10s / per-run cost), 15, 101)` from the slower build's warm-up, so sub-second benchmarks get the power they need while the long ones stay at the floor. Under that rule this benchmark would have been sampled 50 times and the false reading could not have occurred. Worth carrying into the Docker and Windows baselines.

## 2. #841 nursery sweep — recommends 64m

Master only, 4 sizes × 3 benchmarks × 15 runs, 0 failures.

| size | binarytrees | bench_gc_churn | bench_gc_large_heap |
|---|---|---|---|
| 16m | +35.3% time, -48.8% RSS | -2.0% time, -84.1% RSS | **+318% time, +136% RSS** |
| 32m | +19.5% time, -48.1% RSS | -1.4% time, -72.0% RSS | **+317% time, +192% RSS** |
| **64m** | +6.5% time, -26.2% RSS | -0.7% time, -48.0% RSS | +0.7% time, same RSS |
| 128m | baseline | baseline | baseline |

64m is the only size that is never bad. 16m and 32m are disqualified by a **mechanism**, not a preference: on a large live set they are 4.2x slower *and* use more memory, because the live set stops fitting the nursery and survivors get promoted — 28 MB promoted against zero at 64m. That transfers to other machines in a way the +6.5% does not.

Churn's p95 pause scales with nursery size (665us at 16m → 4759us at 128m) — the throughput-versus-latency trade, for a future release that targets pause.

**This is one machine and one architecture.** The size that fits a live set is workload-dependent by definition, so the Windows and Docker sweeps should agree before 64m becomes the default.

## 3. #861 — reproduced on arm64, fix verified

macOS has no per-process affinity, so the pinned Linux configuration is unreachable and the sequential loop there had no power at all. Oversubscription reproduces the same starvation using the fixture's own constants (`ceil(cpus * 3 / 8)` = 6 copies on 14 cores, ~48 threads).

| arm | mode | runs | crashes | rate |
|---|---|---|---|---|
| pre-fix `ad2c610bb0` | jit=off | 5000 | 4 | 0.08% |
| pre-fix | jit=1 | 5000 | 7 | 0.14% |
| post-fix `e864d335fd` | jit=off | 5000 | **0** | — |
| post-fix | jit=1 | 5000 | **0** | — |

All 11 crashes share one signature: `SIGSEGV KERN_INVALID_ADDRESS at 0x20`, faulting thread `MemoryManager::CheckPdaRoots+460`, with an exiting thread in `UnregisterMutator` and a collection in `CollectMinor ← AllocateObject`. `StackFrame::jit_mem` is at offset 0x20, so every one is `cur_frame->jit_mem` with `cur_frame == nullptr`.

If both arms shared a rate, all 11 landing in one arm has probability 2 × 0.5^11 = 0.00098. Identical bytecode and identical `lang.obl` on both arms; only the VM differs.

**Scope, stated precisely:** this shows the fix holds on macOS arm64. It does **not** demonstrate the `RegisterMutator` synchronisation race or the `*stack_frame = nullptr` hunks — every arm64 crash is the exiting-thread window that the `CheckPdaRoots` null guard closes.

Side effect: the oversubscribed path's detection rate, recorded as unmeasured in 94ab3e25f9, now has a number — 0.11% on 14-core arm64 against 0.26% pinned on Linux x64, so it is a working substitute for pinning rather than a check that merely proves the fixture builds.

## 4. `jit_closure_call_in_creating_frame` on arm64 — 12/12 PASS

s0 and s3 × {`--jit=off`, default threshold, `OBJECK_JIT_THRESHOLD=1`} × {verifier off, on}. Not a hollow pass: `OBJECK_GC_STATS=1` shows `minor=12 major=12 pauses=24 promoted_objects=484`, one collection per `GcBetween` iteration. So 42ac1763ec's "ARM64 mirrored, untested" is now tested.

## Notes for review

- **`perf-results/` is gitignored**, so these files are force-added — the same way the 30 files already tracked there were. Plan item 0.3 asks for "raw samples committed to `perf-results/baseline-2026.10.0/`" with "baselines committed" as its exit criterion; the ignore exists to keep ad-hoc local runs out of the tree.
- **The driver's `stress` mode carries the superseded sequential #861 design** and should be replaced by the oversubscription design above before reuse. The README says so.
- **One caveat on absolute numbers:** Xcode auto-updated two minutes before the baseline started and the load averages were still elevated. ABAB interleaving protects the master-vs-9.4 comparison; absolute figures from early in that pass may be slightly inflated.
- `perf-results/run_benchmarks.sh` does not run on macOS at all — it times with GNU `/usr/bin/time -f`, which BSD time has no flag for. The driver here reads the VM's `[gc-stats]` line instead and records both RSS instruments.


objeck:
Closing as a duplicate of #869, which carries the same data and is the one to merge. Two sessions opened a PR for the same results seven minutes apart, neither aware of the other — my fault for saying I would open it without checking whether the Mac could.

The branches differ in exactly two things:

```
git diff --stat origin/perf/mac-arm64-baselines origin/perf/mac-arm64-baseline-2026.10.0
 perf-results/baseline-2026.10.0/README.md   25 +-
 perf-results/objeck_mac_bench.py            395 ++++++
```

**The driver is why this one closes rather than the other.** `objeck_mac_bench.py` is the pre-fix copy of the same 395 lines already landing as `perf-results/objeck_bench.py`, and it differs in ways that matter to anyone who reuses it:

- its `stress` mode carries the superseded sequential #861 design, which has **no power without CPU affinity** — macOS arm64 has since measured 11 failures in 10,000 pre-fix runs under oversubscription against 0 in 900 sequential on Linux, so that design cannot detect the bug it names. The corrected driver refuses to run the mode rather than emit a clean number with nothing behind it.
- peak RSS is measured on macOS only, falling back elsewhere to the VM's own counter — which does not exist before `0a5e7ba008`, i.e. exactly the v2026.9.4 baseline arm, so a Linux or Windows comparison would silently have no RSS for the "before" build.
- no `.exe` handling, so it cannot locate `obc`/`obr` on Windows at all.

Landing both would leave the uncorrected copy in the tree under a second name, which is precisely how a superseded design outlives its correction.

I checked whether this PR's README had anything worth lifting across and it does not: both carry the Xcode-update caveat and the stress-mode warning, the gitignore justification is in neither (it is in the PR body), and the only differing section is the Driver note — where #869's is the more detailed of the two, since it records which driver produced the numbers and how that version differed.

**No data is lost.** The measurements are byte-identical between the branches; #869 has all eleven data files. Keeping the per-arm summary plus the 11 failure captures rather than the raw 20,000-row CSV is also the right level for the repo — the summary and captures support every claim made from them.

Thanks for the run — the arm64 #861 reproduction is the strongest evidence any fix in this release carries.

PR #879, opened 2026-09-16
vm: a runtime error or Runtime->Exit with a live thread no longer crashes (#877)

### LINKED ISSUE #877: VM: a Nil dereference in JIT code exits 139 with a core on linux-arm64 but 1 elsewhere
## Symptom

A Nil dereference in JIT code **does not exit the same way on every platform**. From nightly run 35027973357 (verify-major leg), as recorded under "Parked findings" in `docs/PLAN_2026_10_0_HARDENING.md`:

- **linux-arm64:** `https_persistence_test` printed `>>> Attempting to dereference a 'Nil' memory instance in native JIT code <<<` in `System.String->Equals`, then **exited 139 with a core dump**.
- **linux-x64 and macOS arm64:** same diagnostic, then **exit 1**.

The underlying Nil dereference in that test was fixed separately. What remains open is that **on linux-arm64 the VM's own error path crashed after reporting the error**, instead of exiting cleanly.

## Why the suite can't see it

`EXPECT_RUNTIME_ERROR` accepts **any** non-zero exit, so exit 139 with a core passes a negative test just as exit 1 does. A crash in the error-reporting path is invisible to the regression runner.

## New data: macOS arm64 does not reproduce it

Checked on an Apple M4 Max against master `e0af3bc0b2`, running the fixtures the plan names at `-opt s0`/`s3` under `--jit=off`/`--jit=1`:

| fixture | exit, all four configurations | crash reports |
|---|---|---|
| `bad_runtime_null` | 1 | 0 |
| `jit_nil_inlined_field` | 1 | 0 |
| `jit_nil_inlined_field_float` | 1 | 0 |
| `jit_nil_inlined_field_funcref` | 1 | 0 |
| `jit_nil_inlined_field_object` | 1 | 0 |

The diagnostic text differs by mode as expected ("... in native JIT code", "... in JIT-to-JIT call"), but the exit code is 1 in all 20 cells.

**What this supports:** the 139-plus-core exit is **not universal to ARM64**. macOS arm64 exits cleanly, so the difference is specific to linux-arm64 (or to that run). It says nothing about the cause.

## Next steps

1. **Reproduce on linux-arm64.** A `probe/**` branch runs the stress-probe workflow on `ubuntu-24.04-arm`. Loop the five fixtures above plus `https_persistence_test`, and capture the core with `gdb` backtraces for all threads (the existing `probe/816-core*` branches show how).
2. **Make the difference visible to the suite.** Give negative tests an expected exit code (for example `# EXPECT_RUNTIME_ERROR: exit=1`), or treat signal exits (128+n, core) as failures, so a crashing error path stops passing silently.

objeck:
Fixes #877. **This isn't specific to linux-arm64.** A runtime error or `Runtime->Exit` crashes deterministically on Linux x64, Linux arm64 and Windows x64 whenever **another thread is running**. The program prints its error and then dies with SIGSEGV: exit 139 instead of 1, or instead of the `Runtime->Exit` status.

## How it was found

The first cloud probe (run 35161446656, linux-arm64) ran the five single-threaded Nil fixtures: 160/160 exits were 1, with no cores. But the original crash was a Nil dereference *inside a multithreaded TLS server*. A threaded fixture reproduced it on the first run, and a control isolated the cause:

| peer threads running | Nil dereference, Main or worker, `--jit=off` and `--jit=1` |
|---|---|
| 0 | exit 1, every run |
| 1 or 6 | **exit 139, every run** (Windows x64, Linux x64, Linux arm64) |

`Runtime->Exit(3)`, divide by zero and index out of bounds all crash the same way. Only a normal return from Main is unaffected: it already halts and drains the other threads (`vm.cpp`).

## Root cause

Those paths call `exit()`, which runs the process's teardown **on the calling thread while the others keep executing**. Two teardown steps crash, each confirmed:

1. **Static destructors free the collector's containers.** Linux x64 core (probe run 35163292474): the main thread is in `__run_exit_handlers` → `~unordered_set` (`MemoryManager::old_generation`), and the peer thread faults in `_M_insert_unique` called from `MemoryManager::AllocateArray`.
2. **Windows' static C++ runtime (`/MT`) frees its locale state at exit.** A first version made the VM's containers never-destroyed. That fixed Linux in every mode but not Windows, and a bisect by peer workload showed why:

| peer workload | never-destroyed containers only (Windows x64) |
|---|---|
| arithmetic, allocation, string concat | 30/30 clean |
| `Int->ToString` (a `std::wstringstream`) | 5/10 crash |
| `Float->ToString` | 10/10 crash |

## Fix

`VmExit(status)` replaces the 156 `exit()` calls on the runtime paths: interpreter, dispatch, traps, collector, verifier and both JITs.

- **When another mutator is registered:** it flushes the C and C++ streams, prints the `OBJECK_GC_STATS` line if enabled, and calls `std::_Exit`. `FatalCollectorOutOfMemory` already exits this way.
- **When no other thread is registered:** it calls `exit()` exactly as before, atexit handlers included. **Single-threaded programs are unchanged.**

Startup exits (loader, argument parsing) still call `exit()`, since no VM thread exists yet. The 44 `exit()` calls in native libraries under `core/lib` are out of scope.

**Trade-off:** in a multithreaded program that exits this way, atexit handlers and static destructors don't run. They were the crash. Output written through C stdio or the C++ streams is flushed first. Before this change it was lost: the test's `PASS` line never reached the pipe on a pre-fix obr.

## Test

`exit_with_live_threads.obs`: a thread calls `Runtime->Exit(0)` while two others allocate objects and arrays and convert Ints and Floats to strings, and Main is blocked joining it.

A negative test couldn't guard this, because the runner accepts any non-zero exit, a crash included. With `Runtime->Exit(0)`, a crash becomes a failure.

| Windows x64 | s0 off | s0 jit1 | s3 off | s3 jit1 |
|---|---|---|---|---|
| pre-fix obr (master VM) | 139 ×5 | 139 ×5 | 139 ×5 | 139 ×5 |
| this branch | 0 + PASS ×5 | 0 + PASS ×5 | 0 + PASS ×5 | 0 + PASS ×5 |

Probe fixtures on Windows x64 with this branch: **110/110 clean exits**. That covers:
- Nil dereference on Main and on a worker with 0, 1 and 6 peers
- `Runtime->Exit`, divide by zero and bounds with a live peer
- every peer workload in the bisect above

Still to run before this leaves draft: the full local regression suite in both JIT modes, and the probe matrix on Linux x64/arm64 against this PR's CI build.


objeck:
## Validation complete

**CI:** green on all five platforms, and the perf gate passes. That includes `exit_with_live_threads.obs` in the regression suite everywhere.

**Linux probe against this PR's CI build** (run 35166524977, `diag/877-nil-exit-arm64`): 32 cells × 3 runs × 2 arches = **192 runs, all with the expected exit code, 0 signal exits, no cores** on ubuntu-24.04 (x64) and ubuntu-24.04-arm.

The cells:
- Nil dereference on Main and on a worker with 0, 1 and 6 busy peers
- `Runtime->Exit(3)`, divide by zero and bounds with a live peer, plus a normal return
- a peer doing arithmetic, allocation, concat, Int `ToString` and Float `ToString`
- the new regression test
- each in `--jit=off` and `--jit=1`

On master's build (run 35163292474), every live-peer cell was 139 with a core on both arches.

**Local Windows x64, full suite in both JIT modes:** 280 passed, 4 skipped, 11 failed, the same set in both modes. All 11 are native-library tests failing to load their DLL (error 126: ml ×6, opencv, sdl/gl, diags ×2, odbc). This temp-dir worktree's deploy doesn't contain `libobjk_ml.dll`, `libobjk_sdl.dll` and the rest, or their SDL2/onnx/opencv dependencies. That's an environment gap, not this change, and CI's Windows legs ran those tests green.

PR #883, opened 2026-09-17
gc: trace a closure's captures even when a conservative scan marked the block (#881)

### LINKED ISSUE #881: Pipeline failing: Nightly hardening [1a64f53b77] verify-major closure_enum_array_capture
**Nightly hardening [1a64f53b77] verify-major closure_enum_array_capture** failed on `master` at `3e115db501`.

A new failure from the nightly hardening workflow (heap-verifier violation).

Issue key: `1a64f53b77` (step and test, without the leg)

| Signature | Leg | Step | Configuration | Test | Message | Occurrences |
| --- | --- | --- | --- | --- | --- | --- |
| `61a3f30138` | linux-arm64 | verify-major | gc-threshold=64k verify=1 | closure_enum_array_capture | runtime error (exit 134) | 1 |

**Message:** runtime error (exit 134)

**linux-arm64:** seed none recorded; reduced program none recorded; artifacts `nightly-linux-arm64`, log `linux-arm64/verify-major.log` from https://github.com/objeck/objeck-lang/actions/runs/35180418561

### Output

~~~
>>> gc-verify: B2 violation in collection 10618 (major, after): closure=0xaab69690e430 class='Byte[] or closure block' field=0 value=0xaab69690de20: typed reference is not a live old-generation object <<<
>>> gc-verify: 1 violation(s) in collection 10618; aborting <<<
timeout: the monitored command dumped core
~~~

### Triage

Fix it, or, if it is understood and tracked, add its signature to `tools/fuzz/known.json` so later nightlies list it as known:

~~~json
{
  "id": "61a3f30138",
  "leg": "^linux\\-arm64$",
  "step": "^verify\\-major$",
  "message": "runtime\\ error\\ \\(exit\\ 134\\)",
  "issue": "#<this issue>",
  "note": "<why this is expected>"
}
~~~

Run: https://github.com/objeck/objeck-lang/actions/runs/35180418561

Close this issue once the pipeline is green again; the next failure reopens a new one.

objeck:
Fixes #881. **#880 is a different bug that this does not fix** (see below). Draft until the regression suite and the arm64 loops are in.

## Root cause

All three typed function-reference roots traced a closure's captures only when their **own** `MarkMemory` call set the block's mark bit:
- `CheckJitRoots` FUNC_PARM (`memory.cpp` 1644)
- `CheckMemory` FUNC_PARM (2048)
- the dirty-holder scan (2216)

A conservative scan can mark the same capture block first. `CheckObject` on a stale word finds no class for a `BYTE_ARY` block and marks it without tracing. The stale word can sit in the JIT temp look-back window or on an operand stack. `CheckJitRoots` runs on its own thread next to the interpreter-frame walk. When it wins, the typed walk sees the bit already set and skips the captures, so a capture reachable only through the block is freed.

**Found by:**
- the nightly heap verifier: `verify-major`, B2 on a closure block, linux-arm64 only
- the macOS session, which traced it with gdb on an **unmodified** `-g` build in an arm64 Ubuntu container: frame slot 5 (`FUNC_PARM`) of `Test:TopLevelClosure` holds closure block C, whose only capture is reachable through C. C also sat at `jit_mem[-5]` in the ARM64 temp look-back window of the current JIT frame, which marked it first.

ARM64 scans the 16 words *below* `jit_mem`, where stale native-call words sit, which is why only that leg saw it. Any VM edit that changed LTO's frame layout could hide it: two verifier-only instrumentation patches did.

**In this repro the stale word was the verifier's own.** Memory at `jit_mem[-5..-1]` decodes as a `GcVerifier` Origin `{C, nullptr, L"closure", 0, true}`. An earlier, shallower verify pass left it there, and a later JIT frame reused that stack region without clearing its temp area.
- **The report is true:** the capture was swept while a typed root reached it.
- **The trigger isn't in plain runs:** that conservative root doesn't occur without the verifier.
- **The weakness doesn't depend on where the stale word comes from:** any conservative mark of a capture block suppressed the typed trace, and this fix removes that dependency.

Whether a stale word from ordinary code (interpreter natives, JIT callbacks) triggers it in production is still being measured.

## Fix

**`MarkClosureCaptures`:** marks the block, then CAS-claims a new `GC_TRACED_BIT` (0x8 in `MARKED_FLAG`).
- It returns true for exactly one typed caller per collection, whoever set the mark bit, and only that caller traces the captures.
- The bit also bounds recursion through nested closures.
- Both sweeps clear it together with the mark bit.
- The verifier's B1 header check now requires it clear after a collection.

**Not done:** skipping byte arrays in `CheckObject`. That would free a real `Byte[]` held only in a JIT temp.

## Proof: forced conservative-first ordering

Ordering patch A (local only, never committed) joins the JIT-roots thread before the interpreter-frame walk, so conservative marks always land first. The two A builds are identical apart from this fix. Both were built in an arm64 Ubuntu 24.04 container with GCC 13.3, `-O3 -flto=auto -mcpu=native`. Configuration: `closure_enum_array_capture`, `OBJECK_GC_VERIFY=1`, default JIT, `--gc-threshold=64k`, runs round-robin across builds.

| build | B2 violations |
|---|---|
| master `6ebe83556b` + A | **10/10**, deterministic (collection 401 in 8 of 10) |
| this branch `5d92132687` + A | **0/10** |
| this branch, unmodified | 0/10 |
| master, unmodified (control) | 7/10 |

## #880 is separate

`closure_block_body`'s violation is **B3** (a dirty capture block still refers to a nursery object after collection 1), not the B2 mark race. It reproduces at the same ~1% rate on master, master + A and this branch + A (2/100, 1/100, 1/100), and its message matches #880's CI output. It's a capture-*fixup* problem, and a separate investigation is under way.

## Evidence before the fix (master `3e115db501`)

| linux-arm64, `--gc-threshold=64k` | verifier on, default JIT | verifier on, `--jit=off` | verifier off |
|---|---|---|---|
| `closure_enum_array_capture` | 30/30 violations | 0/6 | 0/10 non-zero |

linux-x64 is 0/430 on the same configurations.

## Gates (all met)
- [x] Deterministic proof: forced conservative-first ordering, 10/10 → 0/10 (above).
- [x] linux-arm64 container build (GCC 13.3, `-O3 -flto`): regression 291/5/0 with the JIT; 290/5/1 with `--jit=off`, the one failure being `jit_entry_compiled`, which its own header says fails without the JIT. VM flag tests 205/0, including the multithreaded GC loops.
- [x] CI green on all five platforms: regression 291-293 passed / 0 failed in both JIT modes, plus flag, debugger, DAP, formatter and LSP suites.
- [x] linux-arm64 loops against this PR's CI build (run 35252067439): `closure_enum_array_capture` **0 B2 in 60 runs** (default JIT and `--jit=1`), where master was 30/30. linux-x64 control 0/60.
- [x] Nightly hardening on this build (run 35252066327): linux-x64, windows-x64, windows-arm64 and macos-arm64 all clean; differential 272 compared, 0 divergent. linux-arm64's **only** remaining failure is #880's B3, which is separate and addressed in #885. #881's signature no longer appears.

Follow-up filed separately: the verifier's Origin structs stay on the stack where the ARM64 look-back scans them, so `OBJECK_GC_VERIFY` adds conservative roots plain runs lack (#885 covers the reporting half).


objeck:
## Handoff: linux-arm64 loops for this PR (Surface, native WSL2 aarch64)

The macOS session that traced #880/#881 went offline. This session, on the Windows ARM64 Surface (WSL2 Ubuntu 24.04, aarch64), has taken over the checklist item **"linux-arm64 loops"** and the regression part of **"Full regression ... on this build"**. This is report-only: no pushes, and the draft status is unchanged.

**Builds**, both local with `deploy_posix.sh arm64`, not /mnt/c:
- **before** = `6ebe83556b`, this PR's merge-base, so the fix is the only difference. The evidence table above used `3e115db501`.
- **after** = PR head `5d92132`

**Matrix per tree**, `obc -opt s3 -lib cipher,collect,xml,json`:
| test | config | runs |
|---|---|---|
| closure_enum_array_capture | `--gc-threshold=64k`, `OBJECK_GC_VERIFY=1`, default JIT | 10 |
| closure_block_body | 64k + verify, default JIT | 200 |
| closure_block_body | 64k + verify, `--jit=1` | 100 |
| closure_block_body | 64k + verify, `--jit=off` (control) | 50 |
| closure_block_body | 64k, verifier **off** | 100 |

After that, `run_regression.sh arm64` on the **after** tree, default and `OBJECK_JIT_THRESHOLD=1`.

**Not covered here:** the forced conservative-first ordering proof (Mac container), the VM flag tests, the nightly hardening run, and the Origin-struct follow-up issue.

**Progress** as of 2026-09-17 17:13 UTC:
```
== build before @ 6ebe83556 17:09:12
   built ok 17:11:16
== build after @ 5d9213268 17:11:19
   built ok 17:13:43
== loops start 17:13:43
```

The before/after table will follow as a new comment when the run finishes.

objeck:
## linux-arm64 loop results: the fix holds

Follow-up to the handoff above. Native WSL2 Ubuntu 24.04 aarch64 (Snapdragon X1E80100, 12 cores). Both trees were built locally with `deploy_posix.sh arm64`, and the tests were compiled `-opt s3 -lib cipher,collect,xml,json`.
- **before** = `6ebe83556b`, this PR's merge-base
- **after** = `5d92132`, the PR head

| test | config | before | after |
|---|---|---|---|
| **closure_enum_array_capture** | `--gc-threshold=64k`, `OBJECK_GC_VERIFY=1`, default JIT | **10/10 violations** | **0/10** |
| closure_block_body | 64k + verify, default JIT | 0/200 | 0/200 |
| closure_block_body | 64k + verify, `--jit=1` | 0/100 | 0/100 |
| closure_block_body | 64k + verify, `--jit=off` | 0/50 | 0/50 |
| closure_block_body | 64k, verifier off | 0/100 | 0/100 |

"Violations" means runs that printed `>>> gc-verify: B`. In every cell the non-zero exit count equalled the violation count; no crashes or timeouts outside the verifier.

### Notes
1. **The deterministic case flips from always failing to always passing.** Before the fix, `closure_enum_array_capture` aborted on the verifier in every run, matching the 30/30 in the evidence table. After the fix it passed all 10.
2. **Timing agrees.** Before: 10 runs in 30 s total, each aborting at the first violation, about 3 s in. After: 10 runs in 665 s, about 66 s each, which is the full workload. That matches the nightly's measured ~80 s for this test at 64k.
3. **`closure_block_body` does not reproduce on this local build**, before or after (0/450 each), while CI's ubuntu-24.04-arm build showed ~13% (60/460). I checked that it really runs: about 25 ms per run, prints `PASS: lambda block body`, and the .obe is fresh. That fits the PR description's point that LTO frame layout decides whether a stale word lands in the ARM64 look-back window, and a locally built toolchain gives a different layout. **These rows neither prove nor disprove the fix for that test.** Proving it needs loops against this PR's CI linux-arm64 artifact, which remains open.
4. **The verifier-off control is clean on both trees** (0/100). This build gives no evidence of a user-visible failure without the verifier, consistent with the PR's note that the stale root in the repro was the verifier's own `Origin` struct.
5. **Not covered here:** the forced conservative-first ordering proof (Mac container), VM flag tests, the nightly hardening run, and CI-artifact loops. **Regression** (`run_regression.sh arm64`, default and `OBJECK_JIT_THRESHOLD=1`) on the after tree is running and will follow in a short comment.

<details><summary>Raw loop output</summary>

```
RESULT | before | closure_enum_array_capture  | 64k verify jit=default     | runs=10  | violations=10  | nonzero=10  | 30s
RESULT | before | closure_block_body          | 64k verify jit=default     | runs=200 | violations=0   | nonzero=0   | 5s
RESULT | before | closure_block_body          | 64k verify jit=1           | runs=100 | violations=0   | nonzero=0   | 2s
RESULT | before | closure_block_body          | 64k verify jit=off         | runs=50  | violations=0   | nonzero=0   | 1s
RESULT | before | closure_block_body          | 64k NO verify jit=default  | runs=100 | violations=0   | nonzero=0   | 2s
RESULT | after  | closure_enum_array_capture  | 64k verify jit=default     | runs=10  | violations=0   | nonzero=0   | 665s
RESULT | after  | closure_block_body          | 64k verify jit=default     | runs=200 | violations=0   | nonzero=0   | 5s
RESULT | after  | closure_block_body          | 64k verify jit=1           | runs=100 | violations=0   | nonzero=0   | 2s
RESULT | after  | closure_block_body          | 64k verify jit=off         | runs=50  | violations=0   | nonzero=0   | 1s
RESULT | after  | closure_block_body          | 64k NO verify jit=default  | runs=100 | violations=0   | nonzero=0   | 3s
```
</details>

objeck:
## linux-arm64 regression on this PR's head: clean

`run_regression.sh arm64` on the locally built **after** tree (`5d92132`, native WSL2 aarch64):

| pass | result |
|---|---|
| default JIT | **291 passed, 5 skipped, 0 failed** |
| `OBJECK_JIT_THRESHOLD=1` | **291 passed, 5 skipped, 0 failed** |

### CI-artifact loops: could not run on this machine
I also downloaded the `objeck-linux-arm64` artifacts from CI run 35243949113 (master @ `6ebe83556b`) and CI run 35249214325 (this PR @ `5d92132`). **Neither runs any program on this CPU.** Every run of both trees exits 132 (SIGILL), with `--jit=off` too. gdb shows the fault at `Execute+64: cnth x19`, which is an **SVE** instruction.

Cause: every `Makefile.arm64` builds with `-mcpu=native`, and `ubuntu-24.04-arm` runners have SVE2. This Snapdragon X1E80100 has no SVE, and neither do Apple M-series or Raspberry Pi 4/5. So the CI-artifact loops the checklist asks for can only run on an SVE host, such as GitHub's arm runner. The template in branch `docs/handoff-2026-09-17-pr883` does exactly that.

This is not specific to this PR. **The published v2026.9.4 `objeck-linux-arm64` tarball has the same problem.** Its checksum verifies, but hello world dies with `Illegal instruction` here, and its `obr` contains SVE instructions. It will be tracked separately; details are in the handoff branch.

It may also explain the 0/450 local vs ~13% CI split for `closure_block_body`: local `-mcpu=native` on a non-SVE core produces different codegen and frame layout from CI's.

PR #909, opened 2026-09-19
fix(build): obi and the embedding module get HTTP/2 and HTTP/3, and CI checks obi against obr (#897)

### LINKED ISSUE #897: obi and the embedding module are built without HTTP/2 and HTTP/3
`core/vm/common.cpp` compiles its HTTP/2 and HTTP/3 engines only when `OBJECK_HAS_NGHTTP2`, `OBJECK_HAS_NGTCP2` (POSIX) or `OBJECK_HAS_WINHTTP_H3` (Windows) is defined. Without them it falls back to the `#else` stubs.

The builds that ship `obr` and `obd` define them:
- `core/vm/vs/vm.vcxproj`
- `core/vm/make/Makefile.amd64` and `Makefile.arm64`
- `core/debugger/vs/debugger.vcxproj`

On POSIX, `obd` links the VM's `vm.a`.

The module build defines none of them:
- `core/module/vs/module.vcxproj` compiles its own `common.cpp` with `_MODULE;_NO_HALT;...` and no protocol macros.
- On POSIX, `core/module/make/Makefile.*` builds the VM from `core/vm/make/Makefile.amd64.mod` and `Makefile.arm64.mod`, which don't define them either. `obi`'s link line has no nghttp2 or ngtcp2 libraries.

## Effect

The module is what `obi` links for in-process evaluation, and it's the embedding API. Code run at the `obi` prompt, or by a host embedding Objeck, gets no HTTP/2 or HTTP/3, while the same program under `obr` does. `obi`'s full-screen editor is not affected: its F5 compiles with `obc` and runs `obr` in a subprocess.

## History

The HTTP/3-on-Windows plan made this a stated non-goal ("HTTP/3 lands in `obr.exe` only — `obd`, the REPL and the embedding module keep the `#else` stub, exactly as HTTP/2 does today"). It later listed the gap as still open. `obd` has since been fixed; the module is what remains. The plan is being deleted in the docs cleanup, so this issue is now the record.

## To do

Either:
- give the module build the same macros, include paths and libraries as the VM (nghttp2 and WinHTTP on Windows; nghttp2, ngtcp2 and AWS-LC on POSIX, see `tools/deps/build_quic_deps.sh`), or
- document the limitation where `obi` and the embedding API are described.

Check what `runtime.feature.http2` and `runtime.feature.http3` report inside `obi` either way.

objeck:
Fixes #897.

## The bug, measured

The module — what `obi` links for in-process evaluation, and the embedding API — is a separate compilation of the VM sources with its own macros and link line. It never defined `OBJECK_HAS_NGHTTP2`, `OBJECK_HAS_NGTCP2` or `OBJECK_HAS_WINHTTP_H3`, so `common.cpp` fell back to its `#else` stubs.

Same program, same Windows x64 machine:

| | `http2` | `http3` |
|---|---|---|
| `obr` | 1 | 1 |
| `obi`, before | **0** | **0** |
| `obi`, this PR | **1** | **1** |

## Why it's more than two macros

The module is a **static library**. Once it compiles the protocol engines, its objects reference nghttp2, WinHTTP and the QUIC libraries. So **every executable that links it** must link those too, or it fails with unresolved externals. Each consumer gets exactly what the VM links, copied from the VM's own build files:

| platform | module (macros, include path) | consumers (libraries) |
|---|---|---|
| **Windows** | `module.vcxproj` — both macros, all 4 configs | `repl.vcxproj` (`obi`), `module/example` — `nghttp2.lib;winhttp.lib` |
| **Linux** | `vm/make/Makefile.{amd64,arm64}.mod` — both macros + `-I$(OBJECK_DEPS)/include` | `repl/make/*`, `module/example/make/*` — `$(QUIC_LIBS)` after `-lmbedcrypto` |
| **macOS** | `module.xcodeproj` — both macros + `$(OBJECK_DEPS)/include` | `repl.xcodeproj` — the six QUIC archives, in `vm.xcodeproj`'s order |

**No `core/vm` source changes.**

## The CI check

Proposed in review by Komp, the session coordinating this work: every leg of `ci-build.yml` now runs `check_obi_obr_parity.py`. It runs one program under `obr` and under `obi --file`, and **fails if their `runtime.feature.http2`/`http3` differ**. It checks parity, not presence, so a leg lacking HTTP/3 in both binaries is consistent and passes.

This does two jobs: it verifies this fix at runtime on the legs I can't run locally, and it stops `obi` drifting from `obr` again.

A check like this can pass for the wrong reason, so it closes three ways:

- **A binary that prints nothing** — both outputs must contain a well-formed `http2=<0|1> http3=<0|1>` line, so two empty outputs never "agree".
- **`obi` waiting on input** — stdin is closed and each run has a timeout.
- **A shell mangling the program** — arguments go as a list, and `obi` reads the program from a file. (Passing it inline through PowerShell was mangled into `obi`'s usage text while writing this.)

Tested against the real fixed `obi` and two stand-ins:

| `obi` | result |
|---|---|
| the real, fixed one | **PASS** — both `1/1` |
| stand-in reporting `http2=0 http3=0` (the pre-#897 answer) | **FAIL** — differs |
| stand-in exiting 0 having printed nothing | **FAIL** — no feature line |

## What's verified where

| platform | how |
|---|---|
| **Windows x64** | **locally** — rebuilt module and `obi` link with no unresolved externals; `obi` reports `1/1` |
| **Linux x64** | **locally**, the makefiles only: `make -n` shows the module compile carrying both macros and the include path, and `obi`'s link line ending in the six QUIC archives after `vm.a`. The actual link needs the QUIC dependency build, so **the full build and runtime check come from CI** |
| **macOS arm64, Linux arm64, Windows arm64** | **CI only** — the parity check on each leg |

## Not included

- **MSYS2.** Its VM builds define `OBJECK_HAS_NGHTTP2` only (no HTTP/3), and its `.mod` twins define nothing. It's niche and not in CI; noted on #897.
- **The embedding example isn't built by any deploy script or workflow.** It gets the libraries anyway, since it would otherwise fail to link for anyone building it by hand.


objeck:
I added 6b5fa85249 for the linux-arm64 failure in the new parity step (`obi` aborting with `double free or corruption (out)`).

- **It predates this PR.** Probe #911 ran master plus only `check_obi_obr_parity.py`, and master's `obi` aborted identically. The parity check is simply the first thing that ever ran `obi` on linux-arm64 in CI.
- **Cause:** the arm64 makefiles compiled the two halves of the module with different macros. `lang.o` got `-D_X64`, and the module's VM (`vm.a`) was built without `-D_MODULE`. Both change class layouts: `NativeCode`'s constant pools and inline destructor, and `StackProgram`'s `output_buffer`. Teardown through `lang.o`'s x64 view passed a non-heap pointer to `free()`. amd64 and the macOS Xcode module were already consistent.
- **Verified** in probe #912, which was this branch's code: on linux-arm64 the parity check passes (`http2=1 http3=1` for both). `obi` exits 0 plainly, unbuffered, with `MALLOC_CHECK_=3`, and under gdb, and valgrind shows no invalid read, write or free.

The same commit fixes `core/module/example/make/Makefile.arm64`, which also had `-D_X64` plus an x86-only `-m64`. CI never builds the example.

PR #911, opened 2026-09-19
ci probe (do not merge): does master's obi abort on linux-arm64?

objeck:
**Probe, do not merge.** It will be closed as soon as it reports.

PR #909's new parity step fails on linux-arm64: `obi` aborts with `double free or corruption (out)` (exit -6), while `obr` runs the same program cleanly. This branch is master plus #909's `check_obi_obr_parity.py`, with **none of #909's build changes**. It answers one question: does master's own `obi` abort on linux-arm64?

- If it aborts here too, the crash predates #909 and needs its own issue.
- If it prints `http2=0 http3=0`, the crash comes from #909's linking.

Only the linux-arm64 leg runs. A diagnostic step runs obi plainly, with `MALLOC_CHECK_=3`, under gdb and under valgrind, then ends the job.


objeck:
Result: **master's own obi aborts on linux-arm64**, the same way #909's does:

```
FAIL: obi printed no feature line (exit -6)
    | double free or corruption (out)
```

(job 105968182685; master ba66ed1153 plus only `check_obi_obr_parity.py`)

So the crash predates #909; #909's parity check is simply the first thing that ever ran obi on linux-arm64 in CI. The cause is a build-flag mismatch. `core/module/make/Makefile.arm64` compiles `lang.o` with `-D_X64`, and `core/vm/make/Makefile.arm64.mod` compiles the module's VM without `-D_MODULE`. The two halves disagree about `NativeCode`'s and `StackProgram`'s layouts, and teardown through the x64 view hands `free()` a non-heap pointer. The fix goes into #909 as its own commit, verified by #912.

The gdb and valgrind parts of this probe never ran. `obi --quit` with no program enters the interactive REPL, which never exits on a closed stdin, and the run was cancelled there. Closing; the branch will be deleted.

PR #912, opened 2026-09-19
ci probe (do not merge): backtrace obi's linux-arm64 abort with #909

objeck:
**Probe, do not merge.** It will be closed as soon as it reports.

This is #909's head, with only the linux-arm64 leg. Right after the tree is built, a diagnostic step runs `obi --file <program> --quit` in several ways: plainly, unbuffered, with `MALLOC_CHECK_=3`, under gdb (for a backtrace of the `double free or corruption (out)` abort) and under valgrind (for where the block was allocated and first freed). Then it ends the job. Its companion #911 runs the same diagnostics on master's `obi`.


objeck:
Result: **the arm64 flag fix works.** It is now on #909 as its own commit (6b5fa85249); this branch differed from that commit only in the workflow.

On linux-arm64 (run 35470492785, job 105970432236):

- the parity check passes: `obr: http2=1 http3=1`, `obi: http2=1 http3=1`
- `obi --file <program> --quit` exits 0 and prints its line in every mode: plain, unbuffered, `MALLOC_CHECK_=3`, and under gdb ("exited normally")
- valgrind finds no invalid read, write or free. It reports 38 errors from 2 contexts. The one the probe's filter showed is a pre-existing allocator mismatch in the compiler: `Library::~Library()` (`core/compiler/linker.h:966`) frees `LoadFileBuffer`'s `malloc`'d buffer with `delete[]`. That is unrelated to the abort and will be fixed separately. The other context didn't match the filter and isn't shown.

Before the fix, master's obi aborted the same way (#911), so the fault predates #909. Closing; the branch will be deleted.

Issue #913, opened 2026-09-19
Intermittent heap corruption: SIGSEGV at exit walking StackClass::closure_dclrs (gc_closure_capture_nursery_end, macOS arm64, JIT all)


objeck:
`gc_closure_capture_nursery_end` crashed once with exit 139 on macOS arm64. It ran in "Run regression tests with every method compiled (POSIX)", and 296 of 297 tests passed. That was run 35468902734, job 105966154363, on PR #910. The PR changes only the ML K-means code; this test doesn't load `ml`. The test passes on every recent master run, so the failure is intermittent.

## Where it crashed

`obr`'s crash report (`obr-2026-09-19-210519.ips`, from the run's `regression-logs-macos-arm64` artifact, saved before it expires) shows one thread, at teardown:

```
EXC_BAD_ACCESS / SIGSEGV, KERN_INVALID_ADDRESS at 0x0
  std::__tree_is_left_child / __tree_next_iter          <- iterating a std::map
  StackClass::~StackClass()+340                          <- the closure_dclrs loop
  StackProgram::~StackProgram()
  Loader::~Loader()
  Execute
  objeck_main / main
```

`StackClass::~StackClass` (`core/vm/common.h`) walks `closure_dclrs`, a `std::map<int, std::pair<int, StackDclr**>>` built at load time, and hits a null tree link.

## Reading

- **Not a race.** The process has one thread at the crash, and it is past the program: the output file is empty, which fits stdout still buffered when an exit-time abort hits.
- **The C++ heap was corrupted during the run.** The map's nodes are ordinary `malloc` memory the VM never writes to after loading, and something overwrote them.
- **The test exercises collector writes into closure capture blocks.** Capture blocks are `NEW_FUNC_INST` allocations, always old generation, so each is its own `malloc` block. The test is about the collector relocating "the words of a closure block written since the last collection" (193f076fdd). A write past a capture block's end, or through a stale or interior pointer, lands in a neighbouring `malloc` chunk. A load-time map node is a plausible neighbour.

## Suspects, in order

1. **The nursery-end fixup of closure blocks (193f076fdd).** Check that the word count it relocates comes from the block itself, not a declaration count that could belong to another closure or method.
2. **`MarkClosureCaptures`'s CAS on `GC_TRACED_BIT` (#883).** Check that the pointer it claims is always a real capture-block header, never an address taken from a conservative scan. #816 was exactly this class: a mark bit ORed into memory that wasn't an object.
[...]

PR #914, opened 2026-09-19
PROBE (do not merge): loop gc_closure_capture_nursery_end on macOS arm64 (#913)

objeck:
Diagnostic only; will be closed, not merged. Runs just the macos-arm64 leg and loops the test from #913 under several allocator settings to reproduce the teardown corruption and capture a crash report.


objeck:
Done: it reproduced #913 twice in 150 jit1 runs on macOS arm64, gave the crash reports that identified the corrupted closure_dclrs map, and then verified the fix (0 of 400 jit1 runs with #919). Closing; the branch goes with it.

Issue #658, opened 2026-08-27
GC: residual multithreaded segfault under heavy alloc churn — 1 crash in ~450-500 runs, not reproducible since #579


objeck:
Filed to stop this living only in session notes. **Status: one crash ever observed, across roughly 450–500 runs. Not reproducible on current `master`.** This is a soak item, not an active defect.

## History, so the numbers are not misread

- **2026-07-18:** pure multithreaded allocation (8 threads, heavy young churn, `Vector<Box>`, no diags) segfaulted **9–10 of 40 runs (~17–36%)**, confirmed pre-existing on a clean `master` worktree.
- **2026-07-19, PR #579:** root cause of the bulk found and fixed — an **unlocked JIT code-page allocator** (`PageManager::GetPage`), a process-global singleton with no locking. Several mutator threads JIT-compiling *different* methods concurrently raced the shared `holders` vector (push_back → realloc → use-after-free) and the per-holder bump index, producing overlapping code regions and corrupt machine code. Fixed with a `static std::mutex page_mutex` on both AMD64 (`jit_amd_lp64.cpp:6582`) and ARM64 (`jit_arm_a64.cpp:5371`), both present on `master` today. Gating test: `programs/regression/jit_concurrent_compile.obs`. After the fix: mt_alloc 0/100, previously ~20%.
- **2026-08-08:** on post-#579 `master`, **0/40** at the default threshold; **1/30 (rc=139)** under heavier churn with `obr --gc-threshold=1m`. The small threshold forces frequent collections, which is what surfaced it.
- **2026-08-08/09 follow-up:** two campaigns, roughly 450–500 runs total, **zero further crashes**. Measured bound now under ~0.3%.

Treat as *not reproducible on current master*, **not** as fixed — nothing was changed that targets this specific residual.

## Repro recipe

Eight `Thread` subclasses, each looping: allocate a `Vector<Box>`; 2500 `Box` allocations (an `Int`, an interpolated `String`, an 8-slot `Float[]`); keep every third; sum a field. Compile with `-lib gen_collect,concurrent`. Loop the run 30–100× counting `rc != 0`. Run with `obr --gc-threshold=1m`.

A session-independent capture kit was left at `C:/tmp/gcdumps/` on the maintainer's box: procdump64, both loop scripts, `analyze_dump.py`, `analyze_hang.py`, and the repro `.obe` files.

## Remaining suspects

Both unexcluded, both shared with #657:

[...]

Issue #657, opened 2026-08-27
JIT/GC: unexplained one-off "Error in JIT-to-JIT call" and print-sensitive segfaults (diagnostics in place, not reproduced since 2026-06)


objeck:
Filed to stop this living only in session notes. **Not reproduced since 2026-06-06; diagnostics are in place so the next occurrence is actionable.** Close it if a few more months pass without a sighting.

## The two observations

1. PR #538's first `linux-x64` run died once with

   ```
   >>> Error in JIT-to-JIT call: method='Collection.Vector:IsEmpty:' <<<
   ```

   (exit 1) during API-doc generation. Rerun and all subsequent runs green; never seen on Windows. `status < 0` from `JitRuntime::Execute` means a runtime guard fired inside compiled code, which points at a stale or corrupted instance/array pointer.

2. While debugging the float-equality bug, adding or removing a `wcerr` trace in the JIT **flipped which probe crashed** (g2 vs g3), before the real, deterministic bug was found. That is a second, timing-sensitive failure mode in the noise.

## Mechanics already mapped

The young generation **moves** objects (promotion, forwarding, region reset). Fixup covers op stacks, statics, interpreter `frame->mem`, JIT `jit_mem` declaration slots, and a `JIT_TMP_LOOK_BACK` temp window (`memory.cpp`, `FixupRoots`/`CheckJitRoots`).

Possible holes:

- object pointers held in JIT **registers** across an allocating callback — spill slots are fixed up and registers are restored from them, but not every path has been verified
- inline-expanded callee locals: the `inline_local_offset` region is not walked, though `MAX_INLINE_SIZE=20` limits exposure
- multi-threaded workloads such as the doc tool

## What was already done (PR #538, commit cb28786c2)

75+ adversarial repro attempts on Windows and WSL2 Linux with `--gc-threshold=32k` and `OBJECK_JIT_THRESHOLD=1`, doc-tool workload plus synthetic recursive stress: **zero reproductions**.

Audited clean: frame pool (`jit_mem` nulled on get/release), `jit_mem` lifecycle (prolog-set via `&frame->jit_mem` param), mark/fixup op-stack ranges (both `<= stack_pos`), temp window (`JIT_TMP_LOOK_BACK=16` covers all spill slots).

Shipped instead of a fix:

[...]

PR #919, opened 2026-09-19
GC: GetClosureDeclarations looks up instead of inserting (#913)

### LINKED ISSUE #913: Intermittent heap corruption: SIGSEGV at exit walking StackClass::closure_dclrs (gc_closure_capture_nursery_end, macOS arm64, JIT all)

objeck:
Fixes #913.

## Root cause

`StackClass::GetClosureDeclarations` (`core/vm/common.h`) was:

```cpp
inline std::pair<int, StackDclr**> GetClosureDeclarations(const int id) {
  return closure_dclrs[id];
}
```

`operator[]` **inserts** a default-constructed entry when the id is absent, and an id with no closure declarations is ordinary. The collector calls this for every `FUNC_PARM` it walks, from `CheckStatic`, `CheckStack`, `CheckPdaRoots` and `CheckJitRoots`, which run **concurrently**. Two marking threads inserting into the same `std::map` leave its red-black tree inconsistent.

Nothing fails at the time. The program crashes later, at exit, when `~StackClass` walks that map:

```
SIGSEGV at 0x0
  std::__tree_is_left_child / __tree_next_iter
  StackClass::~StackClass()
  StackProgram::~StackProgram()
  Loader::~Loader()
  Execute / main
```

It needs every method compiled because that is when the JIT root walk feeds the lookup ids that aren't in the map.

## The fix

A `find()`, returning `(0, nullptr)` when the id is absent, which is exactly what the inserted default held. `CheckMemory` and `FixupMemory` then loop over zero declarations, as they already did for inserted defaults, so no caller changes. The method is `const` now. All six callers are in the collector and the heap verifier, and each passes the pair straight through.

`closure_dclrs` was the only map with this hazard: the neighbouring `methods[id]` and `classes[id]` index raw arrays.

## Evidence

| where | before | after |
|---|---|---|
| macOS arm64, `OBJECK_JIT_THRESHOLD=1`, probe #914 | **2 of 150** crashed, same stack both times | to be confirmed by the probe on this commit |
| macOS arm64, plain | 0 of 150 | |
| macOS arm64, stress probe on #914 (independent) | **1 of 200** jit1 runs, exit -11 | |
| Linux x64, 870 runs across modes, glibc heap checks | 0 | 0 in 150 more |

The crash report that started this is preserved from the CI artifact, and #913 has its stack.

## Verification

Probe #914 is rebased onto this commit with its jit1 loop raised to 400 runs on macOS arm64. At the observed rate that expects about five failures without the fix; zero is the result to look for. The stress-probe workflow also runs the closure tests 200 times per mode on this PR.


objeck:
**Verified on macOS arm64.** The probe (#914), rebased onto this commit with its loop raised, ran the test 400 times with `OBJECK_JIT_THRESHOLD=1` and 100 times plain: **0 failures, no crash reports** (run 35472880009).

Same probe, same runner, without the fix: 2 of 150 jit1 runs crashed, and the repo's stress probe hit it independently 1 in 200. At that rate 400 runs expects about five crashes, so zero is the answer we wanted.

objeck:
**Scope: this does not explain #658 or #657, and neither should be closed against it.**

The cause needs two things together: a live `FUNC_PARM` for the collector to walk, and lookup ids absent from `closure_dclrs` (which is what every-method-compiled supplies).

- **#658** (multithreaded GC residual): `core_thread_gc_stress.obs` has no function-typed parameters, so nothing there reaches `closure_dclrs`.
- **#657** (one-off JIT-to-JIT failure during doc generation): that was a JIT guard firing mid-run, not a teardown walk, even though the doc tool does use func-refs.

Credit where due: the repo's own stress probe hit this independently, 1 in 200 jit1 runs, on the PR carrying the diagnostic probe.

